PR #40, opened 2020-04-21
Rethink stack

rafaqz:


rafaqz:
serious refactor os stack and series. behaviour is now more standardised. stacks allways return a `GeoArray` when indexed. Arrays do no have window fields, it introduces too many bugs and inconsistencies. We can wait for using `view` with DiskArrays.jl.

also test and codecov passing

Issue #66, opened 2020-09-15
"Error requiring ArchGDAL from GeoData" after using Circuitscape; using GeoData


vlandau:
Here's the stack trace.  It seems to trigger if I've run `using Circuitscape` prior to `using GeoData`

```
 Warning: Error requiring ArchGDAL from GeoData:
│ LoadError: UndefVarError: RasterDataset not defined
│ Stacktrace:
│  [1] getproperty(::Module, ::Symbol) at ./Base.jl:26
│  [2] top-level scope at /root/.julia/packages/GeoData/MVllE/src/sources/gdal.jl:92
│  [3] include(::Function, ::Module, ::String) at ./Base.jl:380
│  [4] include at ./Base.jl:368 [inlined]
│  [5] include(::String) at /root/.julia/packages/GeoData/MVllE/src/GeoData.jl:1
│  [6] top-level scope at /root/.julia/packages/GeoData/MVllE/src/GeoData.jl:80
│  [7] eval at ./boot.jl:331 [inlined]
│  [8] eval at /root/.julia/packages/GeoData/MVllE/src/GeoData.jl:1 [inlined]
│  [9] (::GeoData.var"#124#133")() at /root/.julia/packages/Requires/qy6zC/src/require.jl:85
│  [10] err(::Any, ::Module, ::String) at /root/.julia/packages/Requires/qy6zC/src/require.jl:42
│  [11] (::GeoData.var"#123#132")() at /root/.julia/packages/Requires/qy6zC/src/require.jl:84
│  [12] withpath(::Any, ::String) at /root/.julia/packages/Requires/qy6zC/src/require.jl:32
│  [13] (::GeoData.var"#122#131")() at /root/.julia/packages/Requires/qy6zC/src/require.jl:83
│  [14] listenpkg(::Any, ::Base.PkgId) at /root/.julia/packages/Requires/qy6zC/src/require.jl:15
│  [15] macro expansion at /root/.julia/packages/Requires/qy6zC/src/require.jl:81 [inlined]
│  [16] __init__() at /root/.julia/packages/GeoData/MVllE/src/GeoData.jl:78
│  [17] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:697
│  [18] _require_from_serialized(::String) at ./loading.jl:749
│  [19] _require(::Base.PkgId) at ./loading.jl:1040
│  [20] require(::Base.PkgId) at ./loading.jl:928
[...]

PR #67, opened 2020-09-16
Add ArchGDAL, HDF5, and NCDatasets compat entries to project.toml

vlandau:
In response to #66


rafaqz:
I guess we could do that, and just not load them. The idea was you don't even have to install all of them (because these are really large binary deps). That's what the situation will be once Pkg has glue packages resolved.

But you would also need to put a line with the UUID in `[deps]`. Just adding the version there wont actually do what you want...
Also we should add HDF5.jl and NCDatasets.jl for consistency. 

I think we probably should actually, and switch back once Pkg has glue packages.

vlandau:
Okay sounds good -- sorry I thought I saw a UUID for ArchGDAL already there. I'll add it now.

rafaqz:
What about HDF5.jl and NCDatasets.jl... we probably can't just do one of the optional deps

vlandau:
Ha :) was just commenting about those two but you beat me to it. I'm not sure the versions you want/need for HDF5.jl and NCDatasets.jl so I didn't add them yet.

rafaqz:
Latest :)

vlandau:
Feel free to scrap this PR if there are more significant changes that you need to make.

rafaqz:
It should be fine. Thanks for the push

vlandau:
Okay added those two -- should be good!

rafaqz:
Sweet! I'll merge when travis approves

rafaqz:
Lol Aqua.jl doesn't like GeoData having "Stale deps". I'll have to turn that check off

PR #73, opened 2020-10-10
Add resample wrapper and methods to convert between ArchGDAL.Dataset and GeoArray

### LINKED ISSUE #68: Wrap gdal warp in a general method
https://discourse.julialang.org/t/help-request-using-archgdal-gdalwarp-for-resampling-a-raster-dataset/47039

@vlandau have a go if you like, ping me with any help you need.

Checkout https://github.com/JuliaGeo/GeoFormatTypes.jl for how crs is wrapped as passed to ArchGDAL.

We could maybe just reuse the existing `reproject` method? But applied to the array so it will reproject both the data and the dim. Not sure if that's the clearest. The gdalwarp docs say it does "mosaicing, reprojecting and warping".

There are already methods to reproject the dimensions here https://github.com/rafaqz/GeoData.jl/blob/master/src/reproject.jl

vlandau:
Closes #68. Still a WIP -- no formal tests or docs yet, but can confirm that it is working for resampling a single-band raster from 1km resolution in a UTM projection (left panel in figure below) to 0.002&deg; in EPSG:4326 (right panel in figure below) :tada:. I think it should work for multi-band rasters. An edge case that would have errors would be if different bands have different NoData values, but I'm not sure if GeoData handles that to start.

I had some issues/trouble converting `ArchGDAL.Dataset`s and `RasterDataset`s to `GDALarray`s, so I just skipped the middle man since the `ArchGDAL.Dataset` is being created in memory anyway. I added a method to convert back and forth between `ArchGDAL.Dataset` and `GeoData.GeoArray`.

@rafaqz I wanted to give you a chance to review what I have so far before proceeding further.
|image | image |
|:-------:|:--------:|

### DIFF
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -19,7 +19,7 @@ export resample
 function resample(A::AbstractGeoArray, resolution::Number;
 				  crs::GeoFormat=crs(A),
                   method::String="near")
-    wkt = convert(String, convert(WellKnownText, proj))
+    wkt = convert(String, convert(WellKnownText, crs))
 
     AG.Dataset(A) do dataset
         AG.gdalwarp([dataset], ["-t_srs", "$(wkt)",
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -1,11 +1,12 @@
-function resample(A::GeoArray, proj::GeoFormat, resolution::Number;
-                  method::String = "near")
+function resample(A::AbstractGeoArray, resolution::Number;
+				  crs::GeoFormat=crs(A), 
+                  method::String="near")
     wkt = convert(String, convert(WellKnownText, proj))
 
     AG.Dataset(A) do dataset
         AG.gdalwarp([dataset], ["-t_srs", "$(wkt)",
-                                      "-tr", "$(resolution)", "$(resolution)",
-                                      "-r", "$(method)"]) do warped
+                                "-tr", "$(resolution)", "$(resolution)",
+                                "-r", "$(method)"]) do warped
             GeoArray(warped)
         end
     end
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -1,0 +1,13 @@
+function resample(A::GeoArray, proj::GeoFormat, resolution::Number;
+                  method::String = "near")
+    wkt = convert(String, convert(WellKnownText, proj))
+
+    AG.Dataset(A) do dataset
+        AG.gdalwarp([dataset], ["-t_srs", "$(wkt)",
+                                      "-te_srs", "$(wkt)",
+                                      "-tr", "$(resolution)", "$(resolution)",
+                                      "-r", "$(method)"]) do warped
+            GeoArray(warped)
+        end
+    end
+end
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -4,7 +4,6 @@ function resample(A::GeoArray, proj::GeoFormat, resolution::Number;
 
     AG.Dataset(A) do dataset
         AG.gdalwarp([dataset], ["-t_srs", "$(wkt)",
-                                      "-te_srs", "$(wkt)",
                                       "-tr", "$(resolution)", "$(resolution)",
                                       "-r", "$(method)"]) do warped
             GeoArray(warped)
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -1,5 +1,23 @@
+export resample
+
+"""
+	resample(A::AbstractGeoArray, resolution::Number;
+			   crs::GeoFormat=crs(A),
+			   method::String="near")
+
+`resample` uses `ArchGDAL.gdalwarp` to resample an `AbstractGeoArray`.
+
+## Arguments
+- `A`: The `AbstractGeoArray` to resample.
+- `resolution`: A `Number` specifying the resolution for the output. If the keyword argument `crs` (described below) is specified, `resolution` must be in units of the `crs`.
+
+## Keyword Arguments
+- `crs`: A `GeoFormatTypes.GeoFormat` specifying an output crs (`A` with be reprojected to `crs` in addition to being resampled). Defaults to `crs(A)`
+- `method`: A `String` specifying the method to use for resampling. Defaults to `"near"` (nearest neighbor resampling). See [resampling method](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r) in the gdalwarp docs for a complete list of possible values.
+
+"""
 function resample(A::AbstractGeoArray, resolution::Number;
-				  crs::GeoFormat=crs(A), 
+				  crs::GeoFormat=crs(A),
                   method::String="near")
     wkt = convert(String, convert(WellKnownText, proj))
 
--- a/src/sources/gdal.jl
+++ b/src/sources/gdal.jl
@@ -408,11 +408,11 @@ function unsafe_ArchGDALdataset(A::AbstractGeoArray)
     nbands = size(A, Band)
 
     dataset = AG.unsafe_create("tmp",
-                               driver = AG.getdriver("MEM"),
-                               width = width,
-                               height = height,
-                               nbands = nbands,
-                               dtype = eltype(A))
+                               driver=AG.getdriver("MEM"),
+                               width=width,
+                               height=height,
+                               nbands=nbands,
+                               dtype=eltype(A))
     # write bands to dataset
     AG.write!(dataset,
               data(permutedims(A, (Lon, Lat, Band))),
--- a/src/sources/gdal.jl
+++ b/src/sources/gdal.jl
@@ -34,15 +34,15 @@ end
 # Array ########################################################################
 
 """
-    GDALarray(filename; 
-              usercrs=nothing, 
-              name="", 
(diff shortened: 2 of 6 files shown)


src/sources/gdal.jl
```diff
@@ -391,3 +391,66 @@ build_geotransform(lat, lon) = begin
...
+                  usercrs = nothing,
+                  dims = dims(AG.RasterDataset(dataset), usercrs),
+                  refdims = (),
+                  name = "",
+                  metadata = metadata(AG.RasterDataset(dataset)),
+                  missingval = missingval(AG.RasterDataset(dataset)))
+    GeoArray(AG.read(dataset), dims, refdims, name, metadata, missingval)
+end
+
+function unsafe_ArchGDALdataset(A::GeoArray)
```
rafaqz, line 405: We can maybe make this `AbstractGeoArray`

src/sources/gdal.jl
```diff
@@ -391,3 +391,66 @@ build_geotransform(lat, lon) = begin
...
+    width = size(A, Lon)
+    height = size(A, Lat)
+    nbands = size(A, Band)
```
rafaqz, line 408: We can't actually be sure that a `GeoArray` has any of these dimensions - especially `Band` but the others as well as it could have been sliced. You can use e.g. `if hasdim(A, Lon)` to check if `A` has a `Lat` dim.

You can make a Dataset from 2d without a Band - check the `write` methods.

The GeoArray may also have other dimensions... I was wondering if it would be cool to just loop over the time/vertical etc dimension somewhere and slice it - then convert each slice with GDAL and cat them together afterwards. But don't worry about that now :)

src/sources/gdal.jl
```diff
@@ -391,3 +391,66 @@ build_geotransform(lat, lon) = begin
...
+    # Set the nodata value
+    miss_val = missingval(A)
+    for i in dims(A, Band).val
+        AG.setnodatavalue!(AG.getband(dataset, i), miss_val)
+    end
+
+    dataset
+end
+
+function AG.Dataset(f::Function, A::GeoArray)
```
rafaqz, line 449: Again we could use AbstractGeoArray here

src/sources/gdal.jl
```diff
@@ -391,3 +391,66 @@ build_geotransform(lat, lon) = begin
     gt[GDAL_NS_RES] = step(lat)
     return gt
 end
+
+function GeoArray(dataset::AG.Dataset, key = nothing;
+                  usercrs = nothing,
```
rafaqz, line 396: I like to (kind of) follow the BlueStyle guide - with no spaces around keyword args `=`.

src/resample.jl
```diff
@@ -0,0 +1,12 @@
+function resample(A::GeoArray, proj::GeoFormat, resolution::Number;
```
rafaqz, line 1: Does the projection always change here? can it be a keyword arg `crs=crs(A)`?

rafaqz (commented):
This is really nice, glad it came together fairly smoothly.

src/resample.jl
rafaqz: I guess it follows that the resolution could also default to the existing resolution, so you could just change the crs

src/sources/gdal.jl
vlandau: In function definitions only or also in function calls?

src/resample.jl
vlandau: I think since it's resample and not reproject it might make sense to require that a resolution be specified?

src/sources/gdal.jl
vlandau: Ahh I see -- is there a way to identify the X dimension, the Y dimension, and the number of bands without relying on dim names?

src/sources/gdal.jl
```diff
@@ -277,9 +277,9 @@ metadata(raster::AG.RasterDataset, args...) = begin
     offset = AG.getoffset(band)
     # norvw = AG.noverview(band)
     units = AG.getunittype(band)
-    path = first(AG.filelist(raster))
+    # path = first(AG.filelist(raster))
     meta = AG.metadata(raster.ds)
-    GDALarrayMetadata(Dict("filepath"=>path, "scale"=>scale, "offset"=>offset, "units"=>units))
+    GDALarrayMetadata(Dict("scale"=>scale, "offset"=>offset, "units"=>units))
```
vlandau, line 282: This could be put back to how it was before since I abandoned efforts to convert between `ArchGDAL.Dataset and `GDALarray`

src/sources/gdal.jl
rafaqz: I think both

src/sources/gdal.jl
rafaqz: They should be Lat/Lon/Band for GeoArray - but a GeoArray might be just a vector after indexing or something and not have them all anymore. ANd other formats like netcdf wont have `Band` - they might just be 2d lat/lon arrays. 

But maybe don't worry about this for now I'll fix it up later - I'm distracting you from the main use case with the edge cases.

src/sources/gdal.jl
vlandau: done

src/resample.jl
```diff
@@ -0,0 +1,31 @@
...
+
+`resample` uses `ArchGDAL.gdalwarp` to resample an `AbstractGeoArray`.
+
+## Arguments
+- `A`: The `AbstractGeoArray` to resample.
+- `resolution`: A `Number` specifying the resolution for the output. If the keyword argument `crs` (described below) is specified, `resolution` must be in units of the `crs`.
+
+## Keyword Arguments
+- `crs`: A `GeoFormatTypes.GeoFormat` specifying an output crs (`A` with be reprojected to `crs` in addition to being resampled). Defaults to `crs(A)`
+- `method`: A `String` specifying the method to use for resampling. Defaults to `"near"` (nearest neighbor resampling). See [resampling method](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r) in the gdalwarp docs for a complete list of possible values.
```
vlandau, line 16: @rafaqz this link does not render when I run `?resample` -- but maybe it will for the Documenter.jl-generated docs? Just wanted to flag this.

rafaqz:
Ok this is looking great. Just needs a few tests!

rafaqz:
Also I was thinking I will add `resample` for `GeoStack` and `GeoSeries` so you can convert whole datasets with one command.

vlandau:
> Were you thinking we could just use the "-te" and "-te-srs" flags in gdalwarp to add new extent bounds?

Yep, exactly! Though it might tricky if the snap raster bounds and input raster bounds don't match. Those edge cases were what prevented me from going for it right away.

For tests, I can think of a couple. 1) check that the resolution of the output matches the resolution specified in the function call and 2) check that the extent bounds are the same for the input and output GeoArrays (allowing error up to +/- the step of each dim).

rafaqz:
Oh right that makes more sense.

vlandau:
> Oh right that makes more sense.

Okay I'll add that test against ArchGDAL too.

> I'm not sure what GDAL will do if you specify "te" outside of the bounds

I'm not 100% sure either. I remember this being an issue in GeoData when subsetting (I know that's a totally different animal though) for another workflow. I had to do something like:
```julia
left_bound = max(left_extent_raster1, left_extent_raster2)
right_bound = min(right_extent_raster1, right_extent_raster2)
lower_bound = max(lower_extent_raster1, lower_extent_raster2)
upper_extent = min(upper_extent_raster1, upper_extent_raster2)
```
[Here's the complete code](https://github.com/csp-inc/VermontTerrestrialPassageTool.jl/blob/main/src/utils.jl#L74-L98) if you want to check it out. Something similar might be needed for setting `-te` in `gdalwarp`.

vlandau:
Also I did use gdalwarp with a snap raster in that same workflow, the extent of the snap (and also mask) raster was just always within the bounds of the input raster. [Here's that code](https://github.com/csp-inc/VermontTerrestrialPassageTool.jl/blob/main/src/utils.jl#L115-L144).

vlandau:
@rafaqz what are you using for data sources for testing? I'll need a .tif to test ArchGDAL output -- should I create a GeoArray, write it to test/data, then read it back in with ArchGDAL?

rafaqz:
I just use some random tif file from somewhere... It's in the gdal.jl test file. I need to source more and do that in an organised way.

vlandau:
@rafaqz tests added! I skipped checking the output extents for now (which is a test I suggested above) -- I just test for the correct output resolution and that the data (array) for the `resample` output matches the output from a "from scratch" resample method

rafaqz:
Awesome! I'll add some more when I have a go at snap. It seems the download is broken somehow, but I can merge this when that passes.

vlandau:
Sweet! Yeah it's something simple with `download` failing. Should have it up and passing soon!

vlandau:
:tada:

rafaqz:
croppedsnap

Looks like gdal just fills the edge :) I'll get a PR sorted tomorrow.

vlandau:
Nice! That makes things super easy!

PR #75, opened 2020-10-15
Snap resample

rafaqz:
Add a method to `resample` that snaps one `AbstractGeoArray` to the crs, bounds and resolution of another `AbstractGeoArray`.

This also fixes a serious bug in `bounds` for gdal based GeoArrays. This is now tested against results from rasterio instead of manually calculated bounds.

Some refactors for gdal.jl are included.

@vlandau if you want to have a look at these changes that would be great :)

Sorry its a little annoying to read from moving things around. Most of the duplicated code from the last PR was put in `gdalsetproperties!` that is used for your MEM and the original GTIFF methods that create datasets.

### DIFF
--- a/src/resample.jl
+++ b/src/resample.jl
@@ -2,29 +2,52 @@ export resample
 
 """
 	resample(A::AbstractGeoArray, resolution::Number;
-			   crs::GeoFormat=crs(A),
-			   method::String="near")
+			 crs::GeoFormat=crs(A), method::String="near")
+	resample(A::AbstractGeoArray, snap::AbstractGeoArray; method::String="near")
 
 `resample` uses `ArchGDAL.gdalwarp` to resample an `AbstractGeoArray`.
 
 ## Arguments
 - `A`: The `AbstractGeoArray` to resample.
-- `resolution`: A `Number` specifying the resolution for the output. If the keyword argument `crs` (described below) is specified, `resolution` must be in units of the `crs`.
+- `resolution`: A `Number` specifying the resolution for the output. 
+  If the keyword argument `crs` (described below) is specified, `resolution` must be in units of the `crs`.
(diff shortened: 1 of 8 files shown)


rafaqz:
Waiting on this to be merged and registered before tests will pass:

https://github.com/rafaqz/DimensionalData.jl/pull/188

rafaqz:
Both methods now work when there are additional dimension like `Ti`, say from a Netcdf - and for stacks and series of multiple files.

src/resample.jl
```diff
@@ -2,29 +2,52 @@ export resample
...
-- `crs`: A `GeoFormatTypes.GeoFormat` specifying an output crs (`A` with be reprojected to `crs` in addition to being resampled). Defaults to `crs(A)`
-- `method`: A `String` specifying the method to use for resampling. Defaults to `"near"` (nearest neighbor resampling). See [resampling method](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r) in the gdalwarp docs for a complete list of possible values.
+- `crs`: A `GeoFormatTypes.GeoFormat` specifying an output crs 
+  (`A` with be reprojected to `crs` in addition to being resampled). Defaults to `crs(A)`
+- `method`: A `String` specifying the method to use for resampling. Defaults to `"near"` 
+  (nearest neighbor resampling). See [resampling method](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r) 
+  in the gdalwarp docs for a complete list of possible values.
 
 """
+resample
```
vlandau, line 25: Should this be deleted?

vlandau:
Looks great!!

src/resample.jl
rafaqz: That's just there for the docstring

PR #85, opened 2020-11-04
dimkeys without bounds attrib

### LINKED ISSUE #84: dimkeys without bounds attrib
I've encountered this a few times, where a netcdf file has some `dimkeys` with and without a `bounds` attrib, which causes an error in [this line](https://github.com/rafaqz/GeoData.jl/blob/ae7cd8f6bfeb2f19adeee28ad6c1b95f82d1f05c/src/sources/ncdatasets.jl#L43). Suggest changing the line to something like:
```julia
 boundskeys = [dataset[k].attrib["bounds"] for k in dimkeys if haskey(dataset[k].attrib, "bounds")] 
```

Mattriks:
## This PR:

- Fixes #84

### DIFF
--- a/src/sources/ncdatasets.jl
+++ b/src/sources/ncdatasets.jl
@@ -40,7 +40,7 @@ nondimkeys(dataset) = begin
     dimkeys = keys(dataset.dim)
     removekeys = if "bnds" in dimkeys
         dimkeys = setdiff(dimkeys, ("bnds",))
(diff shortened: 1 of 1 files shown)


rafaqz:
Thanks. I guess it's not that easy to test.

It would be good to have a suite of all the varieties of netcdf to run over in the tests to pick up some of these problems.

PR #88, opened 2020-11-04
Pretty printing for `AbstractGeoStack`

ali-ramadhan:
Took a stab at defining a prettier `Base.show` method for `AbstractGeoStack`.

Before:

```julia
julia> using NCDatasets, GeoData

julia> ds = NCDstack("double_gyre.nc")
NCDstack{String,Tuple{},Tuple{},NCDstackMetadata{String,Any},Tuple{}}("double_gyre.nc", (), (), NCDstackMetadata{String,Any}(Pair{String,Any}("interval", 172800.0),Pair{String,Any}("Oceananigans", "This file was generated using Oceananigans v0.44.0#ali/named-tuple-netcdf"),Pair{String,Any}("Julia", "This file was generated using Julia Version 1.5.2\nCommit 539f3ce943 (2020-09-23 23:17 UTC)\nPlatform Info:\n  OS: Linux (x86_64-pc-linux-gnu)\n  CPU: Intel(R) Xeon(R) Silver 4214 CPU @ 2.20GHz\n  WORD_SIZE: 64\n  LIBM: libopenlibm\n  LLVM: libLLVM-9.0.1 (ORCJIT, cascadelake)\n  GPU: TITAN V\n"),Pair{String,Any}("output time interval", "Output was saved every 2 days."),Pair{String,Any}("date", "This file was generated on 2020-11-03T20:57:16.256."),Pair{String,Any}("schedule", "TimeInterval")), ())
```

This PR so far:

image

# Two issues

1. Couldn't figure out how to pretty print the stack metadata. Was hoping for something more like

```julia
julia> ds.metadata
NCDstackMetadata{String,Any} with 6 entries:
  "interval"             => 172800.0
  "Oceananigans"         => "This file was generated using Oceananigans v0.44.0#ali/named-tuple-netcdf"
  "Julia"                => "This file was generated using Julia Version 1.5.2\nCommit 539f3ce943 (2020-09-23…
  "output time interval" => "Output was saved every 2 days."
  "date"                 => "This file was generated on 2020-11-03T20:57:16.256."
  "schedule"             => "TimeInterval"
```

2. Something is doing A LOT of allocating. I don't understand the GeoData.jl internals so I probably did something wrong...

```julia
julia> @time Base.show(stdout, ds);
3.202175 seconds (873.53 k allocations: 2.215 GiB, 7.48% gc time)
```

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -21,3 +21,28 @@ Base.show(io::IO, A::DiskGeoArray) = begin
     end
     print(io, "\n  From file: $(filename(A))")
 end
+
+function Base.show(io::IO, stack::AbstractGeoStack)
+    printstyled(io, "$(Base.typename(typeof(stack)))", color=:blue)
+    print(io, " with $(length(keys(stack))) fields: $(stack.filename)\n")
+
+    for var in keys(stack)
+        printstyled(io, " $var", color=:green)
+
+        field_dims = dims(stack, var)
+        if length(field_dims) > 0
+            print(io, " with dimensions ")
+            for (d, dim) in enumerate(field_dims)
+                printstyled(io, "$(GeoData.name(dim))($(length(dim)))", color=:red)
+                d != length(field_dims) && printstyled(io, ", ", color=:red)
+            end
+        end
+        print(io, '\n')
+    end
+
+    n_metadata = length(stack.metadata)
+    if n_metadata > 0
+        print(io, "and $n_metadata metadata entries:\n")
+        show(io, stack.metadata)
+    end
+end
--- a/src/show.jl
+++ b/src/show.jl
@@ -25,7 +25,7 @@ end
 function Base.show(io::IO, stack::AbstractGeoStack)
     printstyled(io, "$(Base.typename(typeof(stack)))", color=:blue)
     print(io, " with $(length(keys(stack))) field(s)")
-    applicable(filename, stack) && print(io, ": $(filename(stack))")
+    stack isa DiskGeoStack && print(io, ": $(filename(stack))")
     print(io, '\n')
 
     for var in keys(stack)
@@ -35,7 +35,8 @@ function Base.show(io::IO, stack::AbstractGeoStack)
         if length(field_dims) > 0
             print(io, " with dimension(s) ")
             for (d, dim) in enumerate(field_dims)
-                printstyled(io, "$(GeoData.name(dim))($(length(dim)))", color=:red)
+                printstyled(io, "$(name(dim))", color=:red)
+                print(io, " [length: $(length(dim))]")
                 d != length(field_dims) && print(io, ", ")
             end
         end
@@ -44,7 +45,7 @@ function Base.show(io::IO, stack::AbstractGeoStack)
 
     n_windows = length(stack.window)
     if n_windows > 0
-        print(io, "and with $n_windows window(s):\n")
+        print(io, "and with window:\n")
         for window in stack.window
             print(io, ' ')
             show(window)
--- a/src/show.jl
+++ b/src/show.jl
@@ -24,20 +24,30 @@ end
 
 function Base.show(io::IO, stack::AbstractGeoStack)
     printstyled(io, "$(Base.typename(typeof(stack)))", color=:blue)
-    print(io, " with $(length(keys(stack))) fields: $(stack.filename)\n")
+    print(io, " with $(length(keys(stack))) field(s): $(stack.filename)\n")
 
     for var in keys(stack)
         printstyled(io, " $var", color=:green)
 
         field_dims = dims(stack, var)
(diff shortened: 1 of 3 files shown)


ali-ramadhan:
Ah yeah should probably add a test. And extra information is a `window` is specified.

rafaqz:
This is great. 

I think the allocation is from `dims(stack[var])` - that's loading the array from disk. Instead I think you can do `dims(stack, var)`.

The metadata should act just like a `Dict`. But I don't think it has a show method. Maybe move this code to a `show.jl` file and we can add methods to show the metadata there too.

Printing the `window` if it exists is a good idea.

rafaqz:
Actually even this will allocate a bit because netcdf dims have a vector index. 

`dims(stack, var)`

We might need a way to show dims without loading the index

ali-ramadhan:
Thanks for the feedback @rafaqz! `dims(stack, var)` is indeed ~90x faster so definitely a big improvement, you don't notice the printing taking a while now.

```
julia> @benchmark show(Base.DevNull(), ds)
BenchmarkTools.Trial: 
  memory estimate:  464.09 KiB
  allocs estimate:  4107
  --------------
  minimum time:     34.489 ms (0.00% GC)
  median time:      36.241 ms (0.00% GC)
  mean time:        37.563 ms (0.07% GC)
  maximum time:     62.816 ms (0.00% GC)
  --------------
  samples:          134
  evals/sample:     1
```

ali-ramadhan:
Alright I moved the `show` method to the new `show.jl` and it now prints any windows.

I hacked in pretty printing for the metadata dict by calling `display` (probably should not be doing this, right?).

Looks better now I think:

image

Not a fan of the red color (you immediately think you have errors when it scrolls past you in the terminal haha) but I kept the color to match how dimensions are colored in the rest of the package.

src/show.jl
```diff
@@ -21,3 +21,42 @@ Base.show(io::IO, A::DiskGeoArray) = begin
     end
     print(io, "\n  From file: $(filename(A))")
 end
+
+function Base.show(io::IO, stack::AbstractGeoStack)
+    printstyled(io, "$(Base.typename(typeof(stack)))", color=:blue)
+    print(io, " with $(length(keys(stack))) field(s)")
+    applicable(filename, stack) && print(io, ": $(filename(stack))")
```
rafaqz, line 28: I think better to use `isa DiskGeoStack`, `applicable` is a bit mysterious as to why/why not it would have a `filename` method

src/show.jl
```diff
@@ -21,3 +21,42 @@ Base.show(io::IO, A::DiskGeoArray) = begin
...
+                printstyled(io, "$(GeoData.name(dim))($(length(dim)))", color=:red)
+                d != length(field_dims) && print(io, ", ")
+            end
+        end
+        print(io, '\n')
+    end
+
+    n_windows = length(stack.window)
+    if n_windows > 0
+        print(io, "and with $n_windows window(s):\n")
```
rafaqz, line 47: This can just be "with window:" - the whole tuple is the window, so its not pluralised

src/show.jl
```diff
@@ -21,3 +21,42 @@ Base.show(io::IO, A::DiskGeoArray) = begin
...
+    print(io, '\n')
+
+    for var in keys(stack)
+        printstyled(io, " $var", color=:green)
+
+        field_dims = dims(stack, var)
+        if length(field_dims) > 0
+            print(io, " with dimension(s) ")
+            for (d, dim) in enumerate(field_dims)
+                printstyled(io, "$(GeoData.name(dim))($(length(dim)))", color=:red)
```
rafaqz, line 38: Lets change the `Dim(length)` display, it looks like a method call, and it's not clear that number means length...

Maybe:
```
Dim [length: 45]
```
Or something like that that looks better. Then you can also have the bracket not be red

src/show.jl
ali-ramadhan: Done.

src/show.jl
ali-ramadhan: Done.

src/show.jl
ali-ramadhan: Done.

ali-ramadhan:
Thanks for the review @rafaqz! Here's how it looks now:

image

ali-ramadhan:
Just improved it a little bit around knowing when to use plural nouns:

image

rafaqz:
Looking good! but sorry I had another idea this might be more succinct:

```julia
with 4 dimensions xC, yC, zF, Time 64x65x62x183
```
How about that?

ali-ramadhan:
That's a good suggestion! Prettied it up a bit so it looks like this now:

image

rafaqz:
Ok looks good to merge now

PR #90, opened 2020-11-06
dont show DiskGeoArray data

rafaqz:
Lazy arrays stored on disk shouldn't be shown in the REPL.


maxfreu:
You include show.jl, but it seems to be missing.

rafaqz:
@maxfrue `show` is fixed and merged. `] add GeoData#master` should work for you.

PR #92, opened 2020-11-07
add OpenGeoArray and do block behaviour

rafaqz:
This adds an function/do-block syntax to all `AbstractGeoArray` objects:

```julia
ga = GDALarray(filepath)
ga() do A
    A[I...] # A is a OpenGeoArray with the disk based object loaded underneath. 
    # ...  multiple things you need to do with the open file
end
# It's automatically closed outside of the do block
```

@maxfreu This will probably perform better. There is some type instability do deal with too, guess it's time to start to `@profile` everything and lock all of this down.

In the case above `parent(A)` is the `RasterDataset` object. But you can index into `A` using lat/lon etc if you need to.

### DIFF
--- a/src/open.jl
+++ b/src/open.jl
@@ -1,0 +1,45 @@
+
+# Used internally to expose open disk files inside a `do` block
+struct OpenGeoArray{T,N,D<:Tuple,R<:Tuple,A,Na<:Symbol,Me,Mi} <: AbstractGeoArray{T,N,D,A}
+    data::A
+    dims::D
+    refdims::R
+    name::Na
+    metadata::Me
+    missingval::Mi
+end
+function OpenGeoArray(A, dims, refdims, name, metadata, missingval)
+    OpenGeoArray{eltype(A),ndims(A),map(typeof,(dims,refdims,A,name,metadata,missingval))...}(
+        A, dims, refdims, name, metadata, missingval
+    )
+end
+function OpenGeoArray(f, A::AbstractGeoArray{T,N}) where {T,N}
+    withsourcedata(A) do source
+        OA = OpenGeoArray(source, dims(A), refdims(A), name(A), metadata(A), missingval(A))
+        f(OA)
+    end
+end
+
+"""
+    open(f, A::AbstractGeoArray)
+
+`Open` is used to open any `AbstractGeoArray` and do multiple operations
+on it in a safe way. It's a shorthand for the unexported `OpenGeoArray`
+constructor.
+
+`f` is a method that accepts a single argument - an `OpenGeoArray` object
+which is just an `AbstractGeoArray` that holds an open disk - based object. 
+Often it will be a `do` block: 
+
+```julia
+ga = GDALarray(filepath)
+Open(ga) do A
+    A[I...] # A is an `OpenGeoArray` wrapping the disk-based object. 
+    # ...  multiple things you need to do with the open file
+end
+```
+
+By using a do block to open file we ensure they are always closed again
+after we finish working with them.
+"""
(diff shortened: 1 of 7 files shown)


maxfreu:
Hi! Sorry for the delay. It works much better now.

```julia
geoimg() do ga
    @time ga[1000:2000, 1000:2000, 2:8];
    @time ga[1000:2000, 1000:2000, 2:8];
    #translate first dim
    @time ga[2000:3000, 1000:2000, 2:8];
    #translate second dim
    @time ga[1000:2000, 2000:3000, 2:8];
end
```
now results in
```julia
#  1.106616 seconds (19 allocations: 13.379 MiB)
#  0.010530 seconds (38 allocations: 13.379 MiB, 63.35% gc time)
#  0.003353 seconds (19 allocations: 13.379 MiB)
#  1.247189 seconds (20 allocations: 13.379 MiB)
```
So accessing the same region or in block direction works faster now.

rafaqz:
Great! How about the syntax? does that make sense to you? 

I hadn't thought about this before so now is the time for feedback, before I merge.

maxfreu:
Well, good question. I think it would be better to keep the name of the function call in front of the do block constant. Maybe sth along the lines of:
```julia
(Open?)GDALarray(fpath) do raster
    ...
end
```
But I didn't look up how far this would conflict existing plumbing.

rafaqz:
Maybe something like this?
```julia
Open(ga) do A
    A[I...] 
end
Open(GDALarray(path)) do A
    A[I...] 
end
```
FileIO is very commonly used and exports `open`, but `Open` as a type is free, and this is actually a constructor for `OpenGeoArray` underneath. But `OpenGeoArray(ga) do A` seems too verbose.

Another option is just to do:
```julia
GDALarray("path") do A
    A[I...] 
end
```

But you lose the generality of running it on an arbitrary disk backed object you pass in, like `ga` above

maxfreu:
I would say it is important to keep the generality, so `Open` might be best. Having it as a type constructor feels weird, but with the underlying structure it makes sense. At least I have no better idea off the top of my head.

rafaqz:
I just wrote it up and it does feel kinda weird. `open` would be better I'm just worrying about it clashing too much.

It is actually a shortening of the `OpenGeoArray` constructor, and the only exported way of actually constructing an `OpenGeoArray` so its close to being a constructor, but that does feel like a hollow excuse for it being a bit confusing.

what about `read` ?

rafaqz:
I think we need `OpenGeoStack` too, for Netcdf and HDF5 files...

rafaqz:
This was hust `Open` on `AbstractGeoArray` for now. `Open` for stacks can come later.

src/open.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+function OpenGeoArray(f, A::AbstractGeoArray{T,N}) where {T,N}
+    withsourcedata(A) do source
+        OA = OpenGeoArray(source, dims(A), refdims(A), name(A), metadata(A), missingval(A))
+        f(OA)
+    end
+end
+
+"""
+    open(f, A::AbstractGeoArray)
```
maxfreu, line 24: Tiny typo here

maxfreu:
Regarding your concerns wrt `Open` and `open`: Now its a bit late, but maybe `open` is also ok. I assume that there are not many people who use FileIO.jl and GeoData at the same time. And as far as I know the worst thing that can happen is that the usage then needs a qualifier. But on the other hand, I can work well with the current state.

Aand I just spotted a tiny typo in open.jl, line 24

Lastly: Thank you for the effort!! :slightly_smiling_face:

PR #97, opened 2020-11-12
Dont show AbstractGeoArray data

### LINKED ISSUE #94: Don't show `GeoArray` data by default?
Following #90 I was wondering whether it made sense to not show `GeoArray` data by default for `DiskGeoArray` and `MemGeoArray`?

Usually if I just type `ds[:speed]` at the REPL I want to see metadata about the `speed` field, but it always prints a huge array that fills the terminal (the specific numbers aren't that useful :P) and I need to scroll up to read the metadata I'm interested in.

As a side note, should showing a `MemGeoArray` take this much memory?

```julia
julia> using GeoData, NCDatasets

julia> ds = NCDstack("double_gyre.nc");

julia> u = ds[:u];

julia> typeof(u) |> supertype
MemGeoArray{Float32,4,Tuple{Dim{:xF,Array{Float64,1},Sampled{Ordered{ForwardIndex,ForwardArray,ForwardRelation},Regular{Float64},Intervals{Center}},NCDdimMetadata{String,Any}},Dim{:yC,Array{Float64,1},Sampled{Ordered{ForwardIndex,ForwardArray,ForwardRelation},Regular{Float64},Intervals{Center}},NCDdimMetadata{String,Any}},Dim{:zC,Array{Float64,1},Sampled{Ordered{ForwardIndex,ForwardArray,ForwardRelation},Regular{Float64},Intervals{Center}},NCDdimMetadata{String,Any}},Ti{Array{Float64,1},Sampled{Ordered{ForwardIndex,ForwardArray,ForwardRelation},Irregular{Tuple{Float64,Float64}},Intervals{Center}},NCDdimMetadata{String,Any}}},Array{Float32,4}}

julia> @benchmark show(u)

<...LOTS OF PRINTING...>

BenchmarkTools.Trial: 
  memory estimate:  7.89 MiB
  allocs estimate:  183289
  --------------
  minimum time:     128.102 ms (0.00% GC)
  median time:      143.703 ms (0.00% GC)
  mean time:        147.575 ms (0.53% GC)
  maximum time:     175.777 ms (4.29% GC)
  --------------
  samples:          34
  evals/sample:     1

```

ali-ramadhan:
Hopefully I did it right.

I just dispatched on `::AbstractGeoArray` and seemed like printing `filename(A)` does not work for `MemGeoArray` so I only printed the filename is `applicable(filename, A)`.

Resolves #94

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -16,5 +16,5 @@ Base.show(io::IO, A::AbstractGeoArray) = begin
             print(io, " ", d, "\n")
         end
     end
-    applicable(filename, A) && print(io, "\n  From file: $(filename(A))")
+    A isa DiskGeoArray && print(io, "\n  From file: $(filename(A))")
 end
--- a/src/show.jl
+++ b/src/show.jl
@@ -1,7 +1,4 @@
-
-# Don't show the data in DiskGeoArray. It 
-# defeats the purpose of loading them lazily.
-Base.show(io::IO, A::DiskGeoArray) = begin
+Base.show(io::IO, A::AbstractGeoArray) = begin
     l = nameof(typeof(A))
     printstyled(io, nameof(typeof(A)); color=:blue)
     if label(A) != ""
@@ -19,5 +16,5 @@ Base.show(io::IO, A::DiskGeoArray) = begin
             print(io, " ", d, "\n")
         end
     end
-    print(io, "\n  From file: $(filename(A))")
+    applicable(filename, A) && print(io, "\n  From file: $(filename(A))")
 end


rafaqz:
Instead of `applicable`, can we is use `isa DiskGeoArray`? they have to define `filename` its the interface.

ali-ramadhan:
Done. I used `applicable` so it would print a filename for any `A` that had one, but `A isa DiskGeoArray` is clearer if that's the only `AbstractGeoArray` we expect to have a filename.

PR #126, opened 2021-01-11
Added haskey method

mauro3:
Added haskey method.

I get an error on local tests, but I think that is unrelated:
```
Cannot access file ~.julia/dev/GeoData/src/../test/data/SMAP_L4_SM_gph_20160101T223000_Vv4011_001.h5
```

(PS: just been using GeoData over the last few days.  It's really good, thanks!)


rafaqz:
Thanks!

I don't understand those compat issues, seems we've hit some kind of dependency hell somehow... nothing has changed here to cause that.

mauro3:
I'm not sure what the deps problem is.  Maybe I can look into it later this week.  +1 to drop tests when the file is not available.

PR #129, opened 2021-01-14
Xyz geoplot

rafaqz:
@ali-ramadhan this PR generalises your previous PR to handle any dimension combinations. It needs to use the master of DimensionalData.jl to work.

There are some plots in the tests, would be great if you can make sure they do what you need.

It might make sense to put VertDim on the Y axis in DimensionalData.jl recipes, so all the multi dimension fallbacks do that too, and we can delete GeoZDim recipes here.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 [compat]
 Adapt = "2, 3.0"
 ArchGDAL = "0.5"
-DimensionalData = "^0.15.2, 0.16"
+DimensionalData = "0.16"
 GeoFormatTypes = "^0.2.1, 0.3"
 HDF5 = "0.14, 0.15"
 Missings = "0.4"
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -1,3 +1,7 @@
+abstract type GeoXDim{T,Mo,Me} <: XDim{T,Mo,Me} end
+abstract type GeoYDim{T,Mo,Me} <: YDim{T,Mo,Me} end
+abstract type GeoZDim{T,Mo,Me} <: ZDim{T,Mo,Me} end
+
 """
     Lon <: XDim <: Dimension
     Lon(val=:)
@@ -13,7 +17,7 @@ val = A[Lon(1)]
 mean(A; dims=Lon)
 ```
 """
-@dim Lon XDim "Longitude"
+@dim Lon GeoXDim "Longitude"
 
 """
     Lat <: YDim <: Dimension
@@ -30,7 +34,7 @@ val = A[Lat(1)]
 mean(A; dims=Lat)
 ```
 """
-@dim Lat YDim "Latitude"
+@dim Lat GeoYDim "Latitude"
 
 """
     Vert <: ZDim <: Dimension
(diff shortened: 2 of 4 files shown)


ali-ramadhan:
Thanks for this PR @rafaqz! 

I'll run a simulation to make sure the GeoData.jl plot recipes all work.

Should GeoData.jl now `export GeoXDim, GeoYDim, GeoZDim`?

> It might make sense to put VertDim on the Y axis in DimensionalData.jl recipes, so all the multi dimension fallbacks do that too, and we can delete GeoZDim recipes here.

I'm not as familiar with DimensionalData.jl but makes sense to me. Happy to play around and test if you make the change.

rafaqz:
Seems I've borked the heatmap somehow. Will fix

rafaqz:
Ok this is better.

ali-ramadhan (commented):
Looks good! My old script looks like it still works and produces movies with GeoData.jl (heatmaps might be sideways and I need to adapt min/max limits but it plots!).

https://user-images.githubusercontent.com/20099589/105787957-7da4b980-5f4d-11eb-85ab-b18ad51bcc51.mp4

https://user-images.githubusercontent.com/20099589/105787960-7f6e7d00-5f4d-11eb-85da-50a4736d852a.mp4

rafaqz:
Heatmaps being sideways could mean I've messed up the dimension order somehow?

This line does all the permuting:
```julia
A -> permutedims(A, DD.commondims(>:, (GeoZDim, GeoYDim, GeoXDim, TimeDim, Dimension), dims(A)))
```

It means permute A by the common dimensions between the tuple of abstract `Dimension` types, and `dims(A)`. Its sorted by the first tuple order and matched with `>:`.

So this is the order of dimensions passed to Plots will be: `(GeoZDim, GeoYDim, GeoXDim, TimeDim, Dimension)` skipping any that don't exist.

I think that's right! but let me know if not. Mostly I was just trying to get something this general and abstract working at all. 

Also I invited you as a collaborator here so I can officially put you as a reviewer on things like this :)

ali-ramadhan:
> So the order of array axes passed to Plots will be: `(GeoZDim, GeoYDim, GeoXDim, TimeDim, Dimension)` skipping any that don't exist.

Ah maybe it would make more sense to use `(GeoXDim, GeoYDim, GeoZDim, TimeDim, Dimension)`? Thinking about xy plots usually y is on the vertical axis, and for xz and yz plots usually z is plotted on the vertical axis.

> Also I invited you as a collaborator here so I can officially put you as a reviewer on things like this :)

Happy to help with this kind of stuff!

rafaqz:
> Thinking about xy plots usually y is on the vertical axis, and for xz and yz plots usually z is plotted on the vertical axis.

I think that is what Plots.jl does with `ZDim, YDim, XDim` axis order :) the array axes and dim axis arguments for the plot are backwards to each other - it's kind of confusing actually.  You'll see in the code that I do `ys, xs = dims(A)` then return
`xs, ys, parent(A)` to Plots.

So Z should be on the vertical axis (there are some demos of this in the tests). And here Lat is on the Y axis:

```julia
using Plots, GeoData
ga2 = GeoArray(ones(91) * (-25:15)', (Lon(0.0:4.0:360.0), Lat(-25.0:1.0:15.0), ); name=:Test)
plot(ga2)
```

example

Nice work on the color balancing too, it's good to have zero as neutral by default.

rafaqz:
Any comments on this @ali-ramadha? If you have a script for examples that aren't working that could be useful. I just don't want to merge this if it doesn't make sense for your use case.

(I was just reading your Oceananigans paper, and thinking it would also be great to implement those cube plots. Not sure how you made them!)

PR #144, opened 2021-03-21
add convenience methods geoarray, stack and series

rafaqz:
This PR should make it easier to construct various geoarrays, stacks and series. The methods `geoarray`, `stack` and `series` guess the backend constructors from the file extension, instead of having to specify them manually.

### DIFF
--- a/src/convenience.jl
+++ b/src/convenience.jl
@@ -86,7 +86,7 @@ Write any [`AbstractGeoStack`](@ref) to file, guessing the backend from the file
 
 Keyword arguments are passed to the `write` method for the backend.
 
-If the source can't be save as a stack-like object, individual array layers will be saved.
+If the source can't be saved as a stack-like object, individual array layers will be saved.
 """
 function Base.write(filename::AbstractString, s::AbstractGeoStack; kw...)
     base, ext = splitext(filename)
--- a/src/convenience.jl
+++ b/src/convenience.jl
@@ -1,0 +1,186 @@
+"""
+    geoarray(filename; kw...) => AbstractGeoArray
+
+Load a file path as `AbstractGeoArray`. 
+
+# Keywords
+
+Passed to the constructor for the file type, and commmonly include:
+
+- `crs`: crs to use instead of the detected crs.
+- `mappedcrs`: CRS format like `EPSG(4326)` used in `Selectors` like `Between` and `At`, and
+    for plotting. Can be any CRS `GeoFormat` from GeoFormatTypes.jl, like `WellKnownText`.
+- `name`: `Symbol` name for the array.
+- `dims`: `Tuple` of `Dimension`s for the array. Detected automatically, but can be passed in.
+- `refdims`: `Tuple of` position `Dimension`s the array was sliced from.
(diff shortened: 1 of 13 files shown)


src/convenience.jl
```diff
@@ -0,0 +1,186 @@
...
+        arrays = map(filenames) do dn
+            _constructor(geoarray, fn)(fn; window=window, child_kwargs...)
+        end
+        GeoStack(arrays; kw...)
+    end
+end
+
+# Support methods
+
+function _constructor(method::Function, filename; throw=true)
```
mauro3, line 136: I'd add a comment here to what this function does and why you do the in-direction.  But I probably use more comments than you do.

src/convenience.jl
```diff
@@ -0,0 +1,186 @@
...
+    series(dirpath::AbstractString, dims; kw...) => AbstractGeoSeries
+
+Load a vector of filepaths as a `AbstractGeoSeries`. `kw` are passed to the constructor.
+
+`dims` Dimensions can hold an index matching the files in the directory,
+or a function to convert the filename strings to index values.
+"""
+function series end
+
+function series(dirnames;
```
mauro3, line 116: I actually never used `series` nor am I really familiar with what their usecase is.  Anyway, this function returns not a GeoSeries but a GeoStack.  Is this intentional?  If so, is it not contradicting the doc-string?

mauro3:
I did a relatively cursory review of this and LGTM.  The only thing to add is to add these prominently to the Readme & Docs.

src/convenience.jl
rafaqz: Yep, I don't think I comment enough either.

src/convenience.jl
rafaqz: Oh right good catch, I'm not sure what I've done here. It should return a series.

Series is useful for lazy loading huge datasets that are split in multiple files over some dimension/s, often time. So you can index it with the time and get a stack/geoarray result.

rafaqz:
Good to get a sanity check when adding large new things that should become the default syntax. These methods are first in the respective docs sections, but I'll use them everywhere in the README as well, thanks for the reminder.

src/convenience.jl
```diff
@@ -0,0 +1,186 @@
...
+
+# Arguments
+
+- `filename`: A `NamedTuple` of stack keys and `String` filenames.
+
+# Keywords
+
+Passed to the constructor for the file type, and commmonly include:
+
+- `window`: A `Tuple` of `Dimension`/`Selector`/indices that will be applied to the
```
mauro3, line 51: Should/could `geoarray` have this keyword too?

rafaqz: It's been a question... for geoarray it kind of doubles up `view`, so it could be better if `view` actually worked on `DiskGeoArray` lazily - but we need another wrapper layer to achieve that. `window` is like a `view` for an uninstantiated object - because `view` has to wrap an `AbstractArray` and we don't have that yet in a stack.

mauro3: Thanks, makes sense.

PR #168, opened 2021-05-15
Refactor using internal wrappers

### LINKED ISSUE #64: Improve performance of GeoStack and GeoSeries by sharing and caching dims
When lazily loading `GeoSeries` the dims of the child files are being loaded/calculated for every file. This should be done once for the first file and copied for the rest. The assumption of the series is that all dims are the same. 

This may break when it comes to attached metadata being different - that may need some subtlety to handle correctly.

`GeoStack` is also not sharing dims between layers in all cases. This is complicated by the fact that not all layers have the same dimensions - but they could choose from a list of shared dimensions, as is done internally in netcdf files.

### LINKED ISSUE #70: Standardise access to multiple single slice arrays and a single multidimensional arrays
This problem comes up with NetCDF files with a time dimension, and similar data in tif that are just single slices in time.

You can make a series that holds multiple pointers to the same netcdf with a different `window` for each, and just point to each tif file separately. But it's not an efficient way to load the netcdf. Possibly using CachedStack would get around this somehow.

There is also the complication of multiple netcdf along a time dimension that also have their own time dimension - like annual files that have daily data. How can this be standardised to be accessed just like separate daily tif files for the same sequences? Indexing with just one `Ti(x)` object for both cases is the goal.

It seems like the solution to this is to fully abstract the child objects. So indexing should be:

```julia
series[Ti(Date(2001, 2, 4), :tmax, Lat(Between(10, 25)), Lon(Between(130, 150))]
```

Instead of :
```julia
series[Ti(Date(2001, 2, 4)][:tmax][Lat(Between(10, 25)), Lon(Between(130, 150))]
```

Multiple calls to `getindex` break the abstraction and mean we don't have control over doing things in either a general or performant way.

The tricky part will be when we need to index with `Ti` twice - with `Contains` for the series and `At` for the array.  These dimensions may need some marker trait to specify that they are `Partial` or something. This only really makes sense for `Sampled` so they can just inherit from `AbstractSampled`? (It's seeming like we need to move to traits instead of inheritance one day)

### LINKED ISSUE #134: NCDstack not lazy loading when selecting on dimensions?
I have a very large (200GB) NetCDF file with two variables (layers).

If I load the file using `NCDarray("myfile.nc")` then I can only access the first layer.

So instead I load using `NCDstack("myfile.nc")` this allows me to have access to both layers in the file.

However when selecting a layer, I immediately get an out-of-memory error:
```julia
stack = NCDstack("myfile.nc")
layer1 = stack[:layer1] #<<<- Out of memory error here
point = layer1[Lat(x), Lon(y)]
```

This also happens when I try to pre-select one specific lat/lon coordinate to limit the scope:
```julia
stack = NCDstack("myfile.nc")
point = stack[Lat(x), Lon(y)] #<<<- Out of memory error here
layer1 = point[:layer1]
```

Note, it works normally without error when using NCDarray instead of NCDstack:
```julia
array = NCDarray("myfile.nc")
point = array[Lat(x), Lon(y)]
# No errors, and data isn't loaded into memory yet (lazy-loaded)
```

Is this a bug? Or is this how NCDstack is supposed to work?

### LINKED ISSUE #136: Be more lazy
Following from #134 it seems that defaults for returning arrays from stacks should be shifted to returning the lazy version where possible, instead of copying to ram and returning a `GeoArray`.

We should also fix the use of `view` on `NCDarray` etc so that we can index them lazily using `view`.

rafaqz:
This PR is a massive refactor of GeoData.jl internals, structuring things how I always wanted to but hadn't worked out how to do in practice initially.

In its final form this should have minimal impact on use, but a large impact on maintainability, clarity and consistency.

Notably, nearly all the code deletions are actually code, and the additions are mostly extra tests. We should have about 800 lines less code in the packege after this.

Notable changes:
- `AbstractGeoStack` is now `<: AbstractDimStack`. This means the Tables.jl interface will work as well as a lot of base array methods applied to all layers, and the packages are consistent again (this was not easy)
- `FileArray/FileStack` are used as internal wrappers instead of raw strings. This is the main thing that allowed removing so much code and standardizing everything further.
- Reads are delayed to the absolute last minute. Indexing a stack gives you a lazy GeoArray that is still a closed file. It will only be opened when you retreive some data.
- `read(A)` can be used to transfer data to memory - for `GeoArray`, `GeoStack` and `GeoSeries` - to manually break laziness. Using `GeoArray(A)` no longer does this - it may still may be lazy internally. 
- `NCDarray/NCDstack/GRDarra/GDALarray` etc types are all gone. We can add aliases to them to keep code working for a while, but they are no longer needed. Everything is just a `GeoArray` or `GeoStack` and file loading is specified by the wrapped object. 
- DiskArrays.jl will be supported as widely as possible.
- Files can now be _written_ to inside an `open` do block. This can use broadcast etc facilitated by DiskArrays.jl
- Test coverage is higher, due to removing so much code.
- `crop`, `extend` methods allow resizing multiple arrays/stacks to match dimensions
- `slice` allows lazily slicing arrays/stacks to series. This is very useful for standardising accross datasets where e.g. the time dimension may be within a file (e.g. netcdf) or between multipole files (e.g. tif). It will also allow distributed out of core processing as the slices are just string pointers to the files with the metadata to open them.

TODO
- [x] speed up DiskArrays in areas it has made things slower
- [ ] benchmark
- [ ] Fix HDF5Utils.jl so that HDF5 DiskArrays code isn't required here for SMAP

Closes #136 
Closes #134
Closes #64
Closes #70


rafaqz:
This is passing and pretty much finished besides some tests for `extend` and `slice`. 

@mauro3, @ali-ramadhan and @vlandau if you have any time to look over the code (there's a lot so I can't ask for a full review), or could just switch to using this branch that would be a big help.

@meggart DiskArrays.jl has been hugely helpful in this PR, thanks for all the work you put into it. I mostly resolved https://github.com/meggart/DiskArrays.jl/issues/28 using Flatten.jl to re-wrap an opened object with the DiskArray wrappers.
https://github.com/rafaqz/GeoData.jl/pull/168/files#diff-e9f2a6b99d3b555edce8867c798cc88f2d6416b1f170225ec159a2af7cf30c8eR137-R144

So `open` can be used to skip any file opening overhead but keep lazy DiskArrays properties everyerher. In regular use I think just opening/closing multiple times is fine for chunked files.

src/aggregate.jl
```diff
@@ -50,25 +50,22 @@ function aggregate(
...
         map(f, keys_nt)
     end
-    return GeoStack(stack; data=data)
+    return GeoStack(arrays)
 end
 # DimensionalData methods
 """
     aggregate(method, src::AbstractDimArray, scale)
 
 Aggregate an `AbstractDimArray` by `scale` using `method`.
```
vlandau, line 64: Are the possible values for `method` documented somewhere?

vlandau:
@rafaqz I'm still skimming through the code, but had a question regarding the laziness of reading/writing. 

When you subset a disk-backed GeoArray, is only the subset brought into memory, or does that load the entire GeoArray into memory, keep it there, then extract the subset? The former would be pretty amazing, especially because it would readily enable tiling and parallelization of certain tasks that could just work on a chunk of the array at a time. Sounds from your description above like this might be the case? Also, would the same work for writing? Can you lazily write chunks to a disk-backed array? 

This question is mostly selfish, as this sort of lazy reading and writing would make it really easy to reduce [Omniscape.jl](github.com/Circuitscape/Omniscape.jl)'s memory demand and improve scalability efficiency of parallelization.

vlandau (approved):
This looks good to me -- I wasn't able to look through all of the code changes extremely thoroughly, but I like the framework you're implementing here. More laziness, including lazy writing, makes this package scale much more nicely to handle large raster datasets. I'm excited about these changes and improvements.

rafaqz:
@vlandau after this change things are pretty lazy. Data is only loaded when you index with `getindex`, use `read` (which reads a geoarray/stack/series to memory), or apply some Base function like `mean` to the array/stack. 

Views are lazy, and even broadcasting is lazy! (both thanks to DiskArrays.jl). Applying `mean` etc lazily could also be done at some stage if people want that.

The odd thing is that the files are not even open yet when these things happen - `FileArray` just holds array size and the filepath string. This is because I and people I work with (and I assume some others) regularly use thousands of files is sequence and if you open them you hit the OS open files limit. But you might want to take a view over the whole lot beforehand, etc. You could even broadcast e.g. multiplications over thousands of unopened files and it should apply when they are loaded. Not that I've tried that yet...

It should also help with chunk based parallelism, as you can take a view of each chunk (I could make a `chunks` method to do this easily) and its basically just sending the filename to the other thread.

When you actually index with `getindex`, DiskArrays.jl should open chunks at a time if there is chunking, and each chunk opens the file and closes it again. Often there is just one chunk so that doesn't matter.

If you want to work with opened files, or write to files in place, you can use `open(A) do` blocks. In this case the internal `FileArray` is replaced with the actual opened file object like `RasterDataset` for GDAL, keeping all the view/broadcast wrappers etc if they exist.

This should write to disk chunk by chunk without loading the whole file:
```julia
open(A; write=true) do O
    O .*= 2
end
```

All needs a lot testing, please try some and see if they work!

rafaqz:
Seems like chunking is accidentally not enabled for GDAL, and if I enable it some things break. Will have a look...

rafaqz:
Working! and squashed a few bugs and performance bottlenecks in the process.

For now it needs this branch of DiskArrays.jl:
`add https://github.com/rafaqz/DiskArrays.jl#fix_constructors`

vlandau:
Awesome @rafaqz! I'm excited to try it out!

mauro3:
Unfortunately, I don't have the bandwidth to look at this.  I will try out the branch later this month when I'll use using geo-datasets again. (if you want to wait that long)  Although my data fits into memory and I don't do lazy stuff (at least not with data...), so maybe less testing for the changes.

Thanks for the big effort here!

rafaqz:
No problem! It will probably be a few weeks before I merge this too.

mauro3:
Cool! :tada:

PR #221, opened 2021-11-06
generic constructors like zeros make GeoArrays

### LINKED ISSUE #216: Add `zeros`, `ones`, `rand`, `fill`, `falses`, and `trues` to dispatch on `Mapped`/`Projected` mode dims
This may need a trait to separate from DimensionalData somehow. The problem will be finding which dim in the tuple has a GeoData specific index.

like `dimsmode` method in DimensionalData that returns `Nothing` normally but could return `GeoArray`. Then we could use that in dispatch on `zeros`, `ones`, `rand`, `fill`, `falses`, and `trues` to get specific array types based on the index type.

Its only a few lines but maybe its a bit complicated conceptually? @mkborregaard do you think its worth doing this to avoid needing the `GeoArray(falses(dims))` wrapping?

rafaqz:
Closes #216

@mkborregaard this turned out to be super simple with a minor change in https://github.com/rafaqz/DimensionalData.jl/pull/333

`falses(dims(A))` and `trues/zeros/ones/fill/rand` will now make a `GeoArray` if the `LookupArray` for any dimension is `<: AbstractProjected`. Which is every X/Y dim made by GeoData.


mkborregaard:
Haha this does indicate that the design is good

rafaqz:
And most of the added code in DD was just `trues/falses`, the constructor generalisation part is like 5 lines...

PR #225, opened 2021-11-17
inpoly -> inpolygon

mkborregaard:
I'm guessing this is just a leftover typo?

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -677,8 +677,8 @@ end
 function inpolygon(points::AbstractVector, poly::GI.AbstractGeometry)
     inpolygon(points, GI.coordinates(poly))
 end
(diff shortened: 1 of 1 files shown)


mkborregaard:
Still doesn't really work for me though:
```julia
julia> pts
24395-element Vector{Tuple{Float64, Float64}}:
 (121.55573999999999, 14.113560000000001)
 (121.55573999999999, 14.113560000000001)
 ⋮
 (25.217, 0.6)

julia> a[1]
Polygon(137 Points)

julia> inpolygon(pts, a[1])
24395×2 BitMatrix:
 0  0
 0  0
 ⋮  
 0  0
```
I think the result is supposed to be a BitVector, right?

rafaqz:
Ahhg yes this needs testing. The matrix is the output of PolygonInbounds.jl. I guess it should return the first column

PR #230, opened 2021-11-19
print the missingval to IO

vlandau:
The missingval itself was not being printed to io  in `show()`, leading to funny printing during things like docs building with Documenter. I think this simple patch should fix it.


rafaqz:
Thanks!

PR #238, opened 2022-01-12
add skipmissing for missingval

### LINKED ISSUE #229: Should summary functions ignore NoData by default?
When using functions like `minimum`, `maximum`, etc, on a Raster, the missingval is included. Should it be excluded by default? Or maybe new methods for each can be added with a kwarg like `ignoremissingval` so the user can decide?

Right now, I'm doing this to ignore the missingval:

```julia
minimum(raster[raster .!= raster.missingval])
```

That workaround can suffice for now, but I thought the topic was worth bringing up. This is something I can possibly handle a PR for if it is desirable.

rafaqz:
Defines `skipmissing` on `Raster` to also skip `missingval`. This should be a lot faster than `skipmissing(replace_missing(raster))`.

Partly resolve #229

### DIFF
--- a/src/Rasters.jl
+++ b/src/Rasters.jl
@@ -97,6 +97,7 @@ const RasterStackOrArray = Union{AbstractRasterStack,AbstractRaster}
 const RasterSeriesOrStack = Union{AbstractRasterSeries,AbstractRasterStack}
 
 include("utils.jl")
+include("skipmissing.jl")
 include("polygon_ops.jl")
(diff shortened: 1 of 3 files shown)


vlandau (approved):
This looks good to me --  I haven't really work with the internals of iterators in Julia, but this looks straightforward and efficiant. Essentially I see you just had to have skipmissing return a new type, then extend the necessary methods to that type so that summarizing functions work. I do like this solution as an end user!

PR #246, opened 2022-02-01
compare missingval by value

maxfreu:
Here comes the PR. This fixes my use case. The drawback is the explicit check for NaN I introduced. This means that NaN values that were accidentally created in the input to skipmissing could be filtered out unintentionally.

By the way: Why not use the builtin generator?
```julia
Base.skipmissing(r::Raster) = (v for v in r if !(ismissing(v) || isnan(v) || v == missingval(r))))
```

### DIFF
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -1,4 +1,8 @@
+"""
+    skipmissing(itr::Raster)
 
+Returns an iterable over the elements in a `Raster` object, skipping any values equal to either the `missingval` or `missing`.
+"""
 function Base.skipmissing(itr::Raster)
     if ismissing(missingval(itr))
         Base.SkipMissing(itr)
@@ -29,7 +33,14 @@ function Base.iterate(itr::SkipMissingVal, state...)
     item, state
 end
 
-_missing(x, itr) = x === missingval(itr) || x === missing
+_missing(x, itr) = ismissing(x) || x == missingval(itr)  # mind the order, as comparison with missing returns missing
+function _missing(x::AbstractFloat, itr)
+    if isnan(missingval(itr))
+        return ismissing(x) || isnan(x)
+    else
+        return ismissing(x) || x == missingval(itr)
+    end
+end
 
 Base.IndexStyle(::Type{<:SkipMissingVal{T}}) where {T} = IndexStyle(T)
 Base.eachindex(itr::SkipMissingVal) =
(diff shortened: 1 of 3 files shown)


rafaqz:
Thanks! 

I dont think we can have an `isnan` check automatically, `skipmissing` doesn't skip `NaN`, and we shouldn't unless `NaN` is the the explicit `missingval`. 

I suggest adding a method for `_missing(x::AbstractFloat, itr)` where you check if `missingval` is `NaN`, and then do the `isnan` check just for that case.

For your other question, the `SkipMissingVal` object is a copy of the `Base.SkipMissing` object. It can e.g. be indexed with `getindex`, has an `eachindex` method, and has a known `eltype` - unlike your generator.

maxfreu:
> For your other question, the `SkipMissingVal` object is a copy of the `Base.SkipMissing` object. It can e.g. be indexed with `getindex`, has an `eachindex` method, and has a known `eltype` - unlike your generator.

Sounds legit.



> I suggest adding a method for `_missing(x::AbstractFloat, itr)` where you check if `missingval` is `NaN`, and then do the `isnan` check just for that case.

You mean like in the last commit?

rafaqz:
Squash away

src/skipmissing.jl
```diff
@@ -1,4 +1,8 @@
+"""
+    skipmissing(itr::Raster)
 
+Iterates over the given raster, skipping missing values, as well as the builtin `missing`.
```
rafaqz, line 4:
```suggestion
Returns an iterable over the elements in a `Raster` object, skipping any values equal to either the `missingval` or `missing`.
```

src/skipmissing.jl
```diff
@@ -29,7 +33,15 @@ function Base.iterate(itr::SkipMissingVal, state...)
     item, state
 end
 
-_missing(x, itr) = x === missingval(itr) || x === missing
+_missing(x, itr) = ismissing(x) || x == missingval(itr)  # mind the order, as comparison with missing returns missing
+
```
rafaqz, line 37:
```suggestion
```

maxfreu: oh, this slipped through

maxfreu:
Should be good to go now I think.

rafaqz:
Seeing you added a doc string for `skipmissing` documenter wants you to add it to the docs

maxfreu:
How do I do that correctly? Just put it somewhere appropriate in the index.md and add a `@ref` + adding it to the list of exported functions?

rafaqz:
You don't need a `@ref`, just include it in a dics block that has things like that in it.

maxfreu:
Now it seems to work :)

PR #253, opened 2022-03-24
add `append` to ncdatasets write

kongdd:
With the parameter `append`, multiple variables can be saved into the same file.


kongdd:
Still have an issue. The `mode` `a` can not overwritte an existing variable, just add new one. 
But this issue also can be ignored and solved by setting `append = false`.

rafaqz:
I guess that means we need to set `append=false` as the default, or this is a breaking change.

Otherwise this is great, just add `append` to the keywords docs and its good to go.

kongdd:
`append=true` is desired for me. But I am not sure for others. So decision on you.

rafaqz:
Currently writing a file a second time will work, but with this change it wont - so existing scripts will break - I know some of mine will!

We should as much as possible try not to break existing scripts with new features... so lets set it to `false`, and you can choose to set it to `true`. We also can revisit this in future if everyone uses `append=true`.

kongdd:
Agree.

rafaqz:
Thanks! Sorry I'm slow on merging Im insanely busy currently.

kongdd:
No worries. I am trying to bump DiskArrays to v0.3. But it not work. The last commit about `DiskArrays` should be removed.
The others scripts have been passed test.

rafaqz:
Yeah there are some problems with the changes to DiskArrays.jl in v0.3 I have been slowly fixing. Hopefully get that finished soon.

Issue #254, opened 2022-03-25
Low efficiency of `RasterDiskArray `


kongdd:
It seems `Rasters.read` in a extremely low efficiency. It is about 6 times slower than NCDatasets.
Any idea how to improve this?

An unperfect, but practical solution by adding a parameter to escape DiskArray: 
<https://github.com/jl-spatial/Rasters.jl/commit/bc121728cffe506c21afa8e35f550a3cb0294f9a>

kongdd:
```julia
using NCDatasets
using Rasters

f = "clusterId_HI_Tmax_(1961-2016).nc"

# option1
nc = NCDataset(f)
@time data = nc["clusterId"][:];

# option2
x = Raster(f, key = "clusterId")
@time data = read(x);
```
The input file is available at [here](https://mail2sysueducn-my.sharepoint.com/:u:/g/personal/kongdd_mail2_sysu_edu_cn/EbdgY0KqW3lEm8CoY87hKdsBCAXfF1QgaPh4QOG3lRN8Fw?e=bzcV6y).

kongdd:
I think add the parameter `lazy` is the simplest solution,if this issue is not the problem of `RasterDiskArray `.

kongdd:
The function `RasterStack` might have the same issue.

kongdd:
Same issue in reading tiff files:

```julia
using Stars # the GeoArrays backends
using Rasters

f = "/mnt/h/global_WB/ET/diagnosed_PML/PMLV2_v016/PMLV2_yearly_G010_v014_2017-01-01.tif"

@time r = read(Raster(f))
  9.392881 seconds (4.11 M allocations: 793.161 MiB, 2.68% gc time, 10.45% compilation time)
3600×1500×7 Raster{Float64,3} with dimensions:
  X Projected range(-180.0, stop=179.9, length=3600) ForwardOrdered Regular Intervals crs: WellKnownText,
  Y Projected range(89.9, stop=-60.0, length=1500) ReverseOrdered Regular Intervals crs: WellKnownText,
  Band Categorical 1:7 ForwardOrdered

@time r1 = Stars.rast(f)
  1.427316 seconds (128 allocations: 292.901 MiB, 1.16% gc time)
[...]

PR #257, opened 2022-03-29
open files before `read`, `modify`, `collect` and `Array`

rafaqz:
Fixes the performance of `read`, `modify`, `collect` and `Array` by using `open` on the file first. This should have always been the implementation.

@kongdd this PR is a huge improvement to load time without needing the `lazy` flag. We might still need it, as there is always some memory overhead to using lazy DiskArrays methods. But this is a start.

Also notice that in your timings in #254 you were including compilation time in the first run. The `Raster` object is a little complicated and takes a few seconds to compile, and GDAL/ArchGDAL also compile in that first run. We can try to fix that too at some stage, but its a separate issue to this one.

For now, the second and following runs are the real load time without compilation.

This PR:
```julia
julia> using Rasters

julia> f = "/home/raf/Downloads/PMLV2_yearly_G010_v014_2017-01-01.tif"
"/home/raf/Downloads/PMLV2_yearly_G010_v014_2017-01-01.tif"

julia> @time r = read(Raster(f)) # First run: compilation + load time
  4.375702 seconds (10.60 M allocations: 1.133 GiB, 9.57% gc time, 74.30% compilation time)

julia> @time r = read(Raster(f)); # Second run - the actual load time
  0.995884 seconds (2.30 k allocations: 576.877 MiB, 0.72% gc time)
```

Master:
```julia

julia> using Rasters

julia> f = "/home/raf/Downloads/PMLV2_yearly_G010_v014_2017-01-01.tif"
"/home/raf/Downloads/PMLV2_yearly_G010_v014_2017-01-01.tif"

julia> @time r = read(Raster(f));
  8.612977 seconds (9.55 M allocations: 1.072 GiB, 2.60% gc time, 32.18% compilation time)

julia> @time r = read(Raster(f));
  5.713882 seconds (1.01 k allocations: 576.819 MiB, 2.00% gc time)
```

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -28,7 +28,7 @@ missingval(A::AbstractRaster) = A.missingval
 filename(A::AbstractRaster) = filename(parent(A))
 filename(A::AbstractArray) = nothing # Fallback
 
-cleanreturn(A::AbstractRaster) = modify(cleanreturn, A)
+cleanreturn(A::AbstractRaster) = rebuild(A, cleanreturn(parent(A)))
 cleanreturn(x) = x
 
 isdisk(A::AbstractRaster) = parent(A) isa DiskArrays.AbstractDiskArray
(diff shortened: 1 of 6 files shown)


kongdd:
Amazing. Small change, big improvement. 
But, actually, I can't figure out why current script improve the performance.

rafaqz:
@kongdd to make this even more confusing for you ;)

The PR now alters `modify` so it happens at a lower level. `modify` is the DimensionalData.jl way of swapping the internal array of an object for something else, a bit like `set` but just for the array data. Its another way of making it easier to deal with these imuutable objects.

PR #274, opened 2022-06-03
Added time dim definition

Balinus:
Encountered those names with the WRF model

Happy to add tests if you point out where I should develop them.


rafaqz:
Ah that fine, not sure we have to test every specific name in the Dict. 

But a method ambiguity has somehow snuck in with some changes in DimensionalData, I'll fix it and run this again.

rafaqz:
Maybe instead of having lower and upper case versions, we should just lower case `dimname` on this line so both will work for all dimensions. I actually thought it was already like that:
https://github.com/rafaqz/Rasters.jl/blob/c1823befd45bc10ac0a12d1e62597e8abb2acf5a/src/sources/ncdatasets.jl#L260

PR #295, opened 2022-08-20
Small fix: Ensure that dates is always defined

jguerber:
Hey,
I got some errors while using `date` as an `AbstractArray`, this small reorganization should fix it.

### DIFF
--- a/src/sources/rasterdatasources.jl
+++ b/src/sources/rasterdatasources.jl
@@ -83,10 +83,11 @@ function RasterSeries(T::Type{<:RasterDataSource}, layers;
         nothing
     end
     datedim = if haskey(values(kw), :date)
-        dates = if values(kw)[:date] isa Tuple
+        if values(kw)[:date] isa Tuple
             dates = RasterDataSources.date_sequence(T, values(kw)[:date])
             Ti(dates; lookup=Sampled(; sampling=Intervals(Start())))
         elseif values(kw)[:date] isa AbstractArray
+            dates = values(kw)[:date]
             Ti(dates; lookup=Sampled(; sampling=Intervals(Start())))
         else
             nothing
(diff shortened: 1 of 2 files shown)


rafaqz:
Thanks! But this needs a test of whatever it is that didn't work before that worksgiven this change. 

Also including your code and the error in the description would help us understand why we need this change

jguerber:
Yeah sorry that was a bit short :sweat_smile:

In RasterSeries construction, for raster data sources that use a `date` argument, using `date` as an `AbstractArray` allows to fetch only certain dates and not the whole sequence of dates in between. However, it seemed a bit broken:
 
```julia
ser = RasterSeries(AWAP, :solar; date = [Date(2001,1,1), Date(2001,1,5)]) # throws UndefVarError: dates not defined
```
This was caused by the structure of the `if` blocks inside `RasterSeries`, the fix ensures that `Ti(dates,...)` is called only in cases where `dates` is defined. Now it works correctly:

```julia
julia>ser = RasterSeries(AWAP, :solar; date = [Date(2001,1,1), Date(2001,1,5)])
2-element RasterSeries{Raster,1} with dimensions: 
  Ti Sampled Date[2001-01-01, 2001-01-05] ForwardOrdered Irregular Points
```

For consistency, i also added "Tuple" in the error message shown when using RasterSeries with unsupported `date` or `month` arguments (e.g. only one `Date` : `RasterSeries(AWAP, :solar; date = Date(2001,1,5))`).

rafaqz:
Great! Can you add that example to the AWAP tests? Try to use files downloaded in other tests to minimise the CI downloads.

jguerber:
Done :)

rafaqz:
Julia 1.8 has broken something, I'll have to fix before this can be merged

rafaqz:
Could you rebase this against master and push again now that CI is fixed?

jguerber:
Sure :)

PR #311, opened 2022-09-16
Support for ASCII files

jguerber:
Hello !

Here is an imperfect proposition for supporting ASCII files using ASCIIrasters. This is clearly not ready yet but with the start of the new semester i won't be able to work on this as regularly as i did this summer :( The core functions for open/write work quite well but i couldn't test all methods and of the few tests i wrote some are already misfunctioning..

I will try to keep updating from time to time, but feel free to add commits or copy some methods if you're interested !

Because ASCIIrasters is still unregistered and because the live version of RDS does not handle ASCII rasters i had to use submodules to try and make this work in CI, but the easiest way to work on this for now is just to make sure that your package development folder has the right repos with the right branches checked out and to forget CI altogether.

### DIFF
--- a/src/sources/ascii.jl
+++ b/src/sources/ascii.jl
@@ -33,7 +33,7 @@ function DD.dims(ap::ASCIIparams, crs=nothing, mappedcrs=nothing)
     xspan = (xbounds[2] - xbounds[1]) / nc
     yspan = (ybounds[2] - ybounds[1]) / nr
 
-    # Not fully implemented yet
+    # Not implemented yet
     xy_metadata = Metadata{ASCIIfile}(Dict())
 
     xindex = LinRange(xbounds[1], xbounds[2] - xspan, nc)
@@ -67,6 +67,7 @@ end
 
 missingval(ap::ASCIIparams) = params(ap)[:nodatavalue]
 
+# no metadata for now
 DD.metadata(ap::ASCIIparams) = Metadata{ASCIIfile}()
 DD.ndims(ap::ASCIIparams) = 2
 
--- a/src/sources/ascii.jl
+++ b/src/sources/ascii.jl
@@ -78,7 +78,7 @@ function Base.size(ap::ASCIIparams)
     params(ap)[:ncols], params(ap)[:nrows]
 end
 
-Base.Array(ap::ASCIIparams) = _open(a -> Array(a), ap)
+Base.Array(ap::ASCIIparams) = _asciigrid(a -> Array(a), ap)
 
 # Array
 function FileArray(ap::ASCIIparams, filename = filename(ap); kw...)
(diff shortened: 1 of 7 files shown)


rafaqz:
This great. What's left to do to get ASCIIrasters.jl registered?

If you don't get things finished here I can push this over the line at some point once ASCIIraaters.jl is available.

src/sources/ascii.jl
```diff
@@ -0,0 +1,190 @@
...
+###########
+
+function _asciigrid(f::Function, ap::Union{ASCIIparams, FileArray}; kw...)
+    _asciigrid(f, filename(ap), eltype(ap), size(ap); kw...)
+end
+
+function _asciigrid(f, filename::AbstractString, T::Type, size::Tuple; write = false, kw...)
+    dat, pars = open(filename, "r") do io
+        dat, pars = ASCIIrasters.read_ascii(filename; lazy = false)
+        # dat is a nr x nc matrix, we want a nc x nr matrix for use
```
rafaqz, line 164: Why flip this? We can just change the dimension order. 

It's good to minimise allocations in case the files are very large.

jguerber: Indeed, thanks !

rafaqz:
Great, let's get it registered! Let me know if you need any help with that.

rafaqz:
Any news on this now that ASCIIrasters.jl is registered?

jguerber:
Hey :) Sorry for not coming here that often...

As i think you've guessed, it would be nice to be able to fetch some ASCII rasters using RasterDataSources.jl so that we can use them as test files for Rasters.jl (when i checked i couldn't find any good online repositories with some demo ascii rasters that we could have used).

I think that we're not far from being able to merge [RasterDataSources.jl#51](https://github.com/EcoJulia/RasterDataSources.jl/pull/51), but the compatibility errors are a bit frightening. See [ASCIIrasters#9](https://github.com/JuliaGeo/ASCIIrasters.jl/pull/9) for a possible solution ?

rafaqz:
Yeah that will fix it, I just registered your PR. Then we can run the RasterDataSources.jl tests again, get that merged, and *THEN* we can use RasterDataSources.jl to test this.

Although I think there is already an ASCII based data source - yes AWAP returns ASCII `.grid` files as well.

jguerber:
Actually it seems like GDAL is already able to handle `.asc` files so this PR is not really needed. Closing.

rafaqz:
It's good to slowly replace GDAL ;)

One day it will be an optional dependency.

jguerber:
Oh ok haha. I'll keep this in mind then :)

rafaqz:
There are downsides to calling C dependencies.

Eventually if we have WASM compilation or similar things Julia will be able to run in a browser or elsewhere - but the C deps wont. 

GDAL is also a huge black box, if you've ever tried to see how GDAL actually does something it isn't easy. In contrast ASCIIrasters.jl may be the simplest raster loading package that exists.

PR #318, opened 2022-09-26
flexibilize _gdalwrite

maxfreu:
Hi!
This is "slightly" breaking. In this PR I made writing gdal files a bit more flexible. Now it's possible to hand over arbitrary creation options and to use drivers for which the extension is not known (like cog).

Uhm, just noticed something is not working as inteded, but you can already have a look :)

### DIFF
--- a/src/sources/gdal.jl
+++ b/src/sources/gdal.jl
@@ -339,19 +339,15 @@ function _gdalwrite(filename, A::AbstractRaster, nbands;
 )
     A = maybe_typemin_as_missingval(filename, A)
     properties = (width=size(A, X()), height=size(A, Y()), nbands=nbands, dtype=eltype(A))
-    if driver isa String
-        gdaldriver = AG.getdriver(driver)
-    else
-        gdaldriver = driver
-    end
+    gdaldriver = driver isa String ? AG.getdriver(driver) : driver
     options_dict = _parsegdaloptions(options)
 
     if !("COMPRESS" in keys(options_dict))
         options_dict["COMPRESS"] = "ZSTD"
     end
-    
+        
     # drivers supporting the gdal Create() method to directly write to disk
-    drivers_supporting_create = ["GTiff", "COG"]  # this should only be a temporary place to put this
+    drivers_supporting_create = ["GTiff", "HDF4", "KEA", "netCDF", "PCIDSK", "Zarr" #=...=#]  # this should only be a temporary place to put this
 
     # the goal is to set write block sizes that correspond to eventually blocked reads
     # creation options are driver dependent
@@ -379,7 +375,7 @@ function _gdalwrite(filename, A::AbstractRaster, nbands;
 
     options_vec = ["$k=$v" for (k,v) in options_dict]
 
-    if driver in drivers_supporting_create
+    if AG.shortname(gdaldriver) in drivers_supporting_create
         AG.create(filename; driver=gdaldriver, properties..., options=options_vec) do dataset
             _gdalsetproperties!(dataset, A)
             rds = AG.RasterDataset(dataset)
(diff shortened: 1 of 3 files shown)


rafaqz:
Thanks! My only nitpick is that it would be better to use keyword arguments than this `Vector{String}`. That really feels like calling a C API. Keywords feel more like Julian style. And shouldn't need to be breaking?

A `Dict` or `NamedTuble` are worse but fine if there a technical reason not to use keywords.

maxfreu:
The main reason why I used the "options" kwarg in that style is that the lower gdal functions consume it as vector of strings. I also had the impression that it's good to have something that corresponds to the options you would give to `gdal_translate`. Of course it would be more julian to have a kwarg for each option, but documenting them would be hell. This way, you can just write it's the gdal creation options as expected by the driver.

maxfreu:
Ah and regarding breaking stuff: I think it would be possible to reintroduce the compress and chunk keyword args, but to be honest I was too lazy to manipulate the vector stuff to obtain the old behaviour, plus it's quite error-prone. But maybe we should think about it. In general, I don't know if this is the best approach, but there are so many possible keywords and they even depend on the drivers...

src/sources/gdal.jl
```diff
@@ -46,7 +46,7 @@ function Base.write(
     _gdalwrite(filename, correctedA, nbands; kw...)
 end
 function Base.write(
-    filename::AbstractString, ::Type{GDALfile}, A::AbstractRaster{T,3}, kw...
+    filename::AbstractString, ::Type{GDALfile}, A::AbstractRaster{T,3}; kw...
```
maxfreu, line 49: This is a tiny bug fix by the way.

rafaqz: yeah, noted. I think that path just doesn't get hit somehow

rafaqz:
Rasters.jl is pretty solidly about _hiding_ the needs of the lower-level packages!

Is there a reason we cant list all the options and check them? how many are we talking about?

Its pretty easy to automate this kind of thing:
See: https://github.com/rafaqz/GBIF2.jl/blob/main/src/enums.jl

Getting error messages from Rasters.jl from bad options is preferable to getting them from GDAL.

maxfreu:
[GTiff](https://gdal.org/drivers/raster/gtiff.html) alone has 31 options with 2-15 choices. COG has 28 with 2-11 choices. The only julification I see is to convert the kwargs into a vector of strings with key=value pairs, but even then you have the hassle of possibly converting true/false to "YES"/"NO" etc. So although I understand the desire to make it more julian, I'd go with the presented approach. The last problem I see is that currently GDAL.jl doesn't error when it encounter invalid options, so you either have to maintain a list of valid choices, which is a nightmare, or we have to make gdal error. This can then possibly be caught and wrapped in a nicer way or so.

rafaqz:
So you're saying even a `Dict` is not ok... sorry I don't get it. 

Users will input these values from _Julia_, where a vector of `"key=value"` strings is a very unusual object. Why is a `Dict` of `"key => "value"` any worse? 

This is already how `warp` works here. It converts a `Dict` to `Vector{String}`.

maxfreu:
Ah sorry, for the confusion. A `Dict{String,String}` for example would be fine (although a bit more to type). As of now the vector is converted to a dict anyway to fiddle with the user-passed options and then converted back for consumption by lower level functions. So if you want it to be a Dict that's no problem. I just wanted to be consistent with ArchGDAL.

I also found a way to check the creation options:
```julia
AG.validate(driver, ["COMPRESS=ASD"])   # false
AG.validate(driver, ["COMPRESS=ZSTD"])  # true
```
So by looping over all args one can give a good error message.

rafaqz:
Great, using `ArchGDAL.validate` is perfect too.

maxfreu:
I'll be on vacation for a week, so ping me afterwards if I forget to continue work on this :)

alex-s-gardner:
Looking forward to this request being merged ;-)

maxfreu:
I made some edits and added a tiny test. This test however fails, as there is no `gdalinfo`, which I somehow expected to be available. Is there a way to make it available or some other way to retrieve the metadata from within julia?

rafaqz:
Yeah gdalinfo is a command line thing. What do you need some ifically?

maxfreu:
I wanted to check whether the output settings actually have an effect by checking the metadata of the written file. Specifically I want to check for compress=deflate and predictor=2. But maybe it's easier to set some block size and check that via julia.

rafaqz:
Thats a good idea. Sorry I'm not sure how to get that info, you may need to dig around in ArchGDAL.jl/GDAL.jl

maxfreu:
Tests now pass. Note however, that after merging this, code that used the old keyword arguments (compress & chunk) will stop working.

maxfreu:
Hmmm it seems like ZSTD is not supported on some devices / older gdal versions. Maybe it's better to switch to deflate or lzw, even though ZSTD is best. What do you think? Or leave it at the default, which is no compression?

rafaqz:
Thanks for getting this finished!

PR #334, opened 2022-11-16
Documentation updates

jbisits:
All changes are related to the documentation. This is a great package that I have been using recently and in doing so came across small errors in the documentation. I did my best to follow the style of the docs but if there is something you do not like feel free to disregard!

The function [`extract`](https://rafaqz.github.io/Rasters.jl/dev/#Rasters.extract) is currently not displayed as a function so I tried to modify so it would be.

The [example](https://rafaqz.github.io/Rasters.jl/dev/#Rasters.crop) in `crop_extend.jl` does not render the figures. There was a file name error (which I corrected) but I am not sure if this will have solved the issue..

Thanks again for all your work with this great package!!

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -1,8 +1,8 @@
 """
-   extract(x, geoms; atol)
+    extract(x, geoms; atol)
 
 Extracts the value of `Raster` or `RasterStack` at given points, returning
-an iterable of `NamedTuple` with properties for `:geometry` and raster or 
+an iterable of `NamedTuple` with properties for `:geometry` and raster or
 stack layer values.
 
 Note that if objects have more dimensions than the length of the point tuples,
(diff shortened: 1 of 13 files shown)


src/methods/extract.jl
```diff
@@ -1,8 +1,8 @@
 """
-   extract(x, geoms; atol)
+    extract(x, geoms; atol)
```
rafaqz, line 2: Is this space needed? Looks like it has 4 already?

rafaqz (approved):
Thanks, this is amazing. I rarely have time to really proof docs  changes properly, as you can tell.

jbisits:
I can take a look at the codecov/patch tomorrow.

rafaqz:
It's OK that's just noise really

PR #336, opened 2022-11-17
add reduce method to rasterize

### LINKED ISSUE #333: rasterize by a specified function and column
Hi there,
I was wondering if the Raster.jl package is able to rasterize by a specified function and using a specific column.
Thanks

rafaqz:
adds `reduce` funciton argument to rasterize, fixes #333

@harithmorgadinho see if this works for you, `] add Rasters#reduce_rasterize` should check out this branch


harithmorgadinho:
@rafaqz thank you for doing this. I just tried running the rasterization with and without touching. The run without touching seems to be fine but with touching, there seems to be an issue with the pacific islands and the extent of the map. Check the attached figure.
Screenshot 2022-11-17 at 13 38 47

rafaqz:
fixed

rafaqz:
@harithmorgadinho this seems to be competitive with [fasterize](https://github.com/ecohealthalliance/fasterize) now ;)

```julia
 using Rasters, BenchmarkTools, Shapefile, Plots
shppath = "/path/to/MAMMALS_TERRESTRIAL_ONLY/MAMMALS_TERRESTRIAL_ONLY.shp"
shptable = Shapefile.Table(shppath)
mammal_count = rasterize(count, shptable; res=1/6, boundary=:center);
plot(mammal_count);
savefig("mammal_count.png");
```

Here (I think?) `count` does what "sum" does in the fasterize readme benchmark. In this PR `sum` is summing a real column of the table or some other vector you pass in as `fill`, `count` is just counting the presence/absence of the polygons without any `fill`.

mammal_count

I'm not sure what machine fasterize is tested on but my laptop this runs faster than their benchmark of the mammals dataset. I will have to get it working in R to really compare:
```julia
julia> @time mammal_count = rasterize(count, shptable; res=1/6, boundary=:center);
  0.641720 seconds (272.59 k allocations: 2.244 GiB)

julia> @time mammal_count_touches = rasterize(count, shptable; res=1/6, boundary=:touches);
  1.195055 seconds (285.05 k allocations: 2.246 GiB)
```

And with `:touches` its nearly as fast

PR #339, opened 2022-11-29
introduce rgbplot

maxfreu:
Hi! I've just put my plotting code into a dedicated `rgbplot` function, which should make plotting rasters as rgb images convenient :) I've also modified the skipmissing stuff, because I had errors when the missingval was `nothing`.


rafaqz:
Amazing! 

One question: is there a reason you don't use a `plot(raster; rgb=true)` ?

There is quite a bit if code duplication here, and the `RGBPlot` wrapper is a little unusual. Using an `rgb` keyword could solve that.

maxfreu:
I’m just not that familiar with writing plot recipes. I can try to change it.

rafaqz:
Yes its a little mysterious, but there is a `plotattributes` object in scope that has the keyword arguments in it:

So something like this should work:
```julia 
isrgb = haskey(plotattributes, :rgb)
```

And
```julia 
bands = =get(plotattributes, :bands, 1:3)
```

lazarusA:
this looks promising 😄 , any change on finishing this ?

maxfreu:
Hi Lazaro! Well, one would have to re-run the tests and as far as I was concerned, this was pretty much finished. But Rafael suggested to use just "plot" with a rgb keyword instead of "rgbplot". I didn't have the time to include the suggested changes and well, I still don't really have.

Issue #356, opened 2022-12-28
Problem detecting a source file type


yvikhlya:
I am working with a general circulation model which uses netCDF files with different weird file name extensions as inputs and outputs (e.g. .nc3 or .nc4 to distinguish between netCDF3/4). These files can't be loaded into Raster/RasterStack, since source type is derived  by a `_sourcetype(...)` function in `src/convenience.jl` from file name extension, which is .nc for netCDF files, and it defaults to GDAL type. When I use a `source=Rasters.NCDfile` keyword, this does not work either, because this keyword does not propagate down into sub-calls of RasterStack(), and `_open(...)` in `src/convenience.jl` again tries to derive source type from `_sourcetype(filename)` and calls GDAL loader instead of netCDF loader. Here is a complete traceback:

```
julia> st=RasterStack("/home/yvikhlia/aogcm/yv_s2s4mom5_1/geosgcm_ocn2d/yv_s2s4mom5_1.geosgcm_ocn2d.monthly.198202.nc4", source=Rasters.NCDfile)
ERROR: MethodError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer
Stacktrace:
  [1] mapreduce_empty(#unused#::typeof(identity), op::Function, T::Type)
    @ Base ./reduce.jl:367
  [2] reduce_empty(op::Base.MappingRF{typeof(identity), typeof(promote_type)}, #unused#::Type{DataType})
    @ Base ./reduce.jl:356
  [3] reduce_empty_iter
    @ ./reduce.jl:379 [inlined]
  [4] mapreduce_empty_iter(f::Function, op::Function, itr::Vector{DataType}, ItrEltype::Base.HasEltype)
    @ Base ./reduce.jl:375
  [5] _mapreduce(f::typeof(identity), op::typeof(promote_type), #unused#::IndexLinear, A::Vector{DataType})
    @ Base ./reduce.jl:427
  [6] _mapreduce_dim
    @ ./reducedim.jl:365 [inlined]
  [7] #mapreduce#765
    @ ./reducedim.jl:357 [inlined]
  [8] mapreduce
    @ ./reducedim.jl:357 [inlined]
  [9] #reduce#767
    @ ./reducedim.jl:406 [inlined]
 [10] reduce
    @ ./reducedim.jl:406 [inlined]
 [11] pixeltype(ds::ArchGDAL.Dataset)
[...]

PR #359, opened 2022-12-29
Fixed specifying source file type manually in RasterStack

yvikhlya:
This PR is supposed to fix https://github.com/rafaqz/Rasters.jl/issues/356


src/convenience.jl
```diff
@@ -17,5 +17,6 @@ _sourcetype(filenames::NamedTuple) = _sourcetype(first(filenames))
 
 # Internal read method
 function _open(f, filename::AbstractString; kw...)
-    _open(f, _sourcetype(filename), filename; kw...)
+    source=get(kw,:source,_sourcetype(filename))
```
rafaqz, line 20: You can just make a keyword argument for `_open(f, filename::Abstractstring; source=_sourcetype(filename), kw...)`

src/create.jl
```diff
@@ -10,7 +10,8 @@ function create(filename::AbstractString, T::Type, dims::Tuple;
     lazy=true, parent=nothing, suffix=nothing, kw...
 )
     filename = _maybe_add_suffix(filename, suffix)
-    create(filename, _sourcetype(filename), T, dims; lazy, kw...)
+    source=get(kw,:source,_sourcetype(filename))
```
rafaqz, line 13: same here, `source` can just be a keyword

src/stack.jl
```diff
@@ -258,10 +258,10 @@ function RasterStack(filename::AbstractString;
         end
         RasterStack(joinpath.(Ref(filename), filenames); keys)
     else
-        st = if haslayers(_sourcetype(filename))
+        st = if haslayers(source)
```
rafaqz, line 261: exactly

src/stack.jl
```diff
@@ -258,10 +258,10 @@ function RasterStack(filename::AbstractString;
...
             crs = defaultcrs(source, crs)
             mappedcrs = defaultmappedcrs(source, mappedcrs)
-            data, field_kw = _open(filename) do ds
+            data, field_kw = _open(filename, source=source) do ds
```
rafaqz, line 264:
```suggestion
            data, field_kw = _open(filename; source) do ds
```

Since I think v1.6 we can just use the keyword directly. Also better to always use `;` rather than `,` to separate keywords.

src/write.jl
```diff
@@ -8,7 +8,8 @@ Keyword arguments are passed to the `write` method for the backend.
 function Base.write(
     filename::AbstractString, A::AbstractRaster; kw...
 )
-    write(filename, _sourcetype(filename), A; kw...)
+    source=get(kw,:source,_sourcetype(filename))
```
rafaqz, line 11: `source=_sourcetype(filename)` can be a keyword argument above

src/write.jl
```diff
@@ -31,9 +32,9 @@ function Base.write(filename::AbstractString, s::AbstractRasterStack; suffix=not
     else
         base = filename
     end
-    T = _sourcetype(filename)
-    if haslayers(T)
-        write(filename, _sourcetype(filename), s; kw...)
+    source = get(kw,:source,_sourcetype(filename))
```
rafaqz, line 35: Again `source` can be an explicit keyword

rafaqz (changes requested):
Thanks for tracking all of those down!

The main change is how you are using `get` on `kw` - we can just add the `source` keyword instead directly in the function keyword arguments.

This PR will also need some tests that specify the source type manually.

src/convenience.jl
yvikhlya: Done.

rafaqz:
Tests?

yvikhlya:
> Tests?

I have no idea how to write tests, but I I'll look into it.

rafaqz:
Tests are crucial, more important than the code in the long term. Otherwise I or someone else might break this functionality again by accident.

You can see some other tests in the tes/gdal.jl and test/ncdatasets.jl - that might be a good place to add tests where you rename a file to something custom and then force it to the right file type.

Do your best anyway! If not I can add them.

yvikhlya:
Can you advise some good tutorial on writing tests in julia? I feel that I need to make myself more familiar with this anyway.

rafaqz:
Maybe https://bkamins.github.io/julialang/2022/04/22/testing.html?

or https://itnext.io/organizing-tests-in-julia-c3de1e43d6ab

yvikhlya:
Ok, making tests turned out to be not so difficult, but I think that they reveal some more bugs. I copy `test/data/*.nc` files to `*.nc4` files and try to create Raster and RasterStack objects from them. First, `Raster()` works without `source=Rasters.NCDfile` keyword, and with this keyword throws 
```
ERROR: MethodError: no method matching _open(::Type{Array}, ::Type{Rasters.NCDfile}, ::ArchGDAL.RasterDataset{Float32, ArchGDAL.Dataset}; key=nothing)
```
This is inconsistent with RasterStack interface. Second, both Raster and RasterStack created from *.nc files are not equal to Raster and RasterStack created from *.nc4 files, they do not pass isequal test. Is such behavior expected?

rafaqz:
> but I think that they reveal some more bugs

Thats great. Can you push the tests so I can try to find your problems?

yvikhlya:
Not sure if I am doing the right thing, but here is my attempt.

PR #361, opened 2023-01-03
Add basic conversions for ocean variables

jbisits:
Converting ocean variables is something I am having to do a lot so I wrote this set of functions and thought it might be a useful addition for Rasters.jl. It exports one function, `convert_ocean_vars`, that will convert depth -> pressure, practical salinity -> absolute salinity, potential temperature -> conservative temperature from a  `RasterStack/Series` as well as compute the in-situ or potential density. Returned is either a `RasterStack/Series` (matches what was passed in). These conversions are done using [GibbsSeaWater.jl](https://github.com/TEOS-10/GibbsSeaWater.jl). There are limitations to the way I have coded it up, so if you can see improvements please let me know. Tests for these functions have been written as well.


jbisits:
I have not yet added an example as I first wanted to see if you thought it worth adding!

rafaqz:
Seem like usefull tolls, but I'm not sure they should be here! 

My concern is it's a bit field specific when everything here is generic to many fields. Maybe there should be a package OceanRasters.jl in JuliaOcean or somewhere that adds these tools?

Second, everything in Rasters.jl uses `broadcast` rather than for loops. I recommend this because it means that lazy larger than memory disk arrays just work (for loops index one cell at a time, slow for disk access) and also means it will work on GPU without change.

Iteration also works with DiskArrays.jl, But for loops don't.

Last, the function names could be spelled out? `depth_to_pressure` is a better name to me.

Anyway, I'll think about this a bit, and Im happy to at least do a proper code review for you.

test/methods.jl
```diff
@@ -268,37 +268,37 @@ end
...
     @test all(mosaic(last, reg1, reg2) .===
               mosaic(last, irreg1, irreg2) .===
-              mosaic(last, exp1, exp2) .=== [0.1 0.2 missing; 
-                                             0.3 1.1 1.2; 
+              mosaic(last, exp1, exp2) .=== [0.1 0.2 missing;
+                                             0.3 1.1 1.2;
                                              missing 1.3 1.4])
 
-    @test all(mosaic(first, [reverse(reg2; dims=Y), reverse(reg1; dims=Y)]) .=== 
-              [missing 0.2 0.1;
```
rafaqz, line 284: Try not to change anything except what you're working on in a PR

src/methods/oceanconversions.jl
```diff
@@ -0,0 +1,227 @@
...
+            θ_profile = θ[X(At(lon)), Y(At(lat)), Z(find_nm)]
+            Θ[i, j, find_nm] = GibbsSeaWater.gsw_ct_from_pt.(Sₐ_profile[find_nm], θ_profile)
+
+        end
+        rs_dims = (lons, lats, z)
+    else
+        for t ∈ time
+            for (i, lon) ∈ enumerate(lons), (j, lat) ∈ enumerate(lats)
+
+                Sₐ_profile = Sₐ[X(At(lon)), Y(At(lat)), Ti(t)]
```
rafaqz, line 142: You can use a `view` for things like this, you don't need a separate copy of the data

src/methods/oceanconversions.jl
```diff
@@ -0,0 +1,227 @@
...
+                find_nm = findall(.!ismissing.(Sₐ_profile))
```
rafaqz, line 143: you can pass an anonymous function to `findall` rather than broadcasting it and allocating intermediate arrays

But I think you can just use the BitArray directly and not use `findall`

jbisits:
>Seem like usefull tolls, but I'm not sure they should be here!
My concern is it's a bit field specific when everything here is generic to many fields. Maybe there should be a package OceanRasters.jl in JuliaOcean or somewhere that adds these tools?

No problem! I agree that it is quite specific, I will have a look over at JuliaOcean and see if there is somewhere these would fit in (and not a problem if not).

>Second, everything in Rasters.jl uses broadcast rather than for loops. I recommend this because it means that lazy larger than memory disk arrays just work (for loops index one cell at a time, slow for disk access) and also means it will work on GPU without change.

Ah right, I should have been more careful about checking how the rest of the package works sorry!

>Last, the function names could be spelled out? `depth_to_pressure` is a better name to me.

I agree with this, and even for my own use I think naming like this is clearer.

>Anyway, I'll think about this a bit, and Im happy to at least do a proper code review for you.

Only if you have the time and inspiration! Thanks very much for your help on this. I will leave the PR open but feel free to close.

src/methods/oceanconversions.jl
```diff
@@ -0,0 +1,227 @@
...
+
+"""
+    function convert_z_to_p(raster::Raster)
+Convert the depth dimension (`Z`) to pressure using `gsw_p_from_z`. Note that pressure
+depends on depth and _latitude_ so the returned pressure is stored as a variable in the
+resulting `Raster` rather than a vertical dimension.
+"""
+function convert_z_to_p(raster::Raster, rs_dims::Tuple)
+
+    lons, lats, z, time = rs_dims
```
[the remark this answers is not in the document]
jbisits: Yes this was an issue I knew would need to be addressed and will change as you suggest to named dimensions.

test/methods.jl
jbisits: That was a complete accident sorry! I should have spotted it.

src/methods/oceanconversions.jl
```diff
@@ -0,0 +1,227 @@
...
+    Θ = convert_θ_to_Θ(θ, Sₐ, rs_dims)
+    converted_vars = (p = p, Sₐ = Sₐ, Θ = Θ)
+    ρ = isnothing(p_ref) ? in_situ_density(Sₐ, Θ, p, rs_dims) :
+                           potential_density(Sₐ, Θ, p_ref, rs_dims)
+    converted_vars = (p = p, Sₐ = Sₐ, Θ = Θ, ρ = ρ)
+
+    return RasterStack(converted_vars, rs_dims)
+
+end
+function convert_ocean_vars(raster_series::RasterSeries, var_names::NamedTuple; p_ref = nothing)
```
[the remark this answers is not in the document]
jbisits: Can do.

src/methods/oceanconversions.jl
```diff
@@ -0,0 +1,227 @@
...
+`gsw_sa_from_sp`.
+"""
+function convert_Sₚ_to_Sₐ(Sₚ::Raster, p::Raster, rs_dims::Tuple)
+
+    lons, lats, z, time = rs_dims
+    Sₐ = similar(Array(Sₚ), Union{Float64, Missing})
+    if isnothing(time)
+        for (i, lon) ∈ enumerate(lons), (j, lat) ∈ enumerate(lats)
+
+            Sₚ_profile = Sₚ[X(At(lon)), Y(At(lat))]
```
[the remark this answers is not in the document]
jbisits: Yep. I was trying to avoid indexing errors by using `lon` and `lat` but it is at the expense of speed.

jbisits:
I was having more of a think about this and I think what I will do is move these conversions to a package called something like OceanRasterConversions.jl rather than a full OceanRasters.jl implementation. This way if anyone (besides me..) ever wanted to use the conversions they could do so by
```julia
using Rasters, OceanRasterConversions
```
then use the ocean variable conversions. I plan to switch from the `for` loops to `broadcast` as you mention above and make the rest of the changes you suggest above. Once I have this all done and the package is getting close to release could I ask you for a review? Mainly to check that the implementation of the methods is closer to the way you do so in Rasters.jl, no problem if you are too busy though. Hopefully I will get to all this not too far in the future.

PR #375, opened 2023-01-18
MODIS datasets need more keyword arguments

jguerber:
Currently, calling e.g. `Raster(MODIS{MOD13Q}, :NDVI; RDS.britanny...)` throws an `UndefKeywordError` because the kwargs specific to MODIS are filtered out by `_filterkw`

This PR allows more kwargs, allowing to call MODIS sources via `Raster` directly.

### DIFF
--- a/src/sources/rasterdatasources.jl
+++ b/src/sources/rasterdatasources.jl
@@ -15,7 +15,7 @@ Load a `RasterDataSource` as a `Raster`. `T` and `layers` are passed to
 - `month`: an `Int` between `1` and `12`, usually for `Climate` datasets
 - `date`: a `DateTime` object, usually for `Weather` datasets.
 - `res`: a `String` resolution, for datasets with multiple resolutions.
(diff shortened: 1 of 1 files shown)


rafaqz:
Thanks! I noticed this the other day.

rafaqz:
The docs are not your fault... I'm stupidly printing the output of a dataset that keeps having observations added to it, in a doctest :sweat_smile:

jguerber:
Something like this ? I feared that it would be a bit too much for the Rasters doc because it's already listed in the RasterDataSources doc, but it's true that the abbreviations need explanation :sweat_smile:

rafaqz:
If you can just split them into separate keywords  like the ones above (`res`, `month` etc) that would be clearer

jguerber:
Indeed :)

rafaqz:
Thanks

rafaqz:
I think we can fix the interface at some point to accept an Extents.extent object to get tiles - but then we will need lazy concatenation which I never finished.

PR #380, opened 2023-01-27
initial draft

lazarusA:
here, the initial setup. Sorry I didn't have more time to finish the whole thing.


lazarusA:
ohh, it fails due to 
```
value = You must set `ENV["RASTERDATASOURCES_PATH"]` to a path in your system
```
@rafaqz what path should we use for CI?

rafaqz:
It's set in docs CI to ".."

https://github.com/rafaqz/Rasters.jl/blob/main/.github/workflows/Documenter.yml#L17

So it should work already for docs

.github/workflows/Documenter.yml
```diff
@@ -1,20 +1,39 @@
...
-      - uses: julia-actions/julia-buildpkg@latest
-      - uses: julia-actions/julia-docdeploy@latest
+      - uses: julia-actions/setup-julia@v1
+      - uses: julia-actions/cache@v1
+        with:
+          cache-registries: "true"
+      - name: Install documentation dependencies
+        run: sudo apt-get update && sudo apt-get install -y xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev
+      - name: Install pkgs dependencies
+        run: xvfb-run -s '-screen 0 1024x768x24' julia --project=docs -e 'using Pkg; Pkg.instantiate(); Pkg.precompile()'
```
asinghvi17, line 28: It doesn't look like GLMakie is used anywhere?  If it isn't, then no need for the xvfb stuff.

src/plot_rasters.jl
```diff
@@ -0,0 +1,114 @@
+function _balance_grid(nplots)
+    ncols = (nplots - 1) ÷ ceil(Int, sqrt(nplots)) + 1
+    nrows = (nplots - 1) ÷ ncols + 1
+    return ncols, nrows
+end
+
+theme_rasters = Theme(
```
asinghvi17, line 7: Ideally, this should be a Function.  Also, is it necessary to change e.g. font, tick size, etc here?

src/plot_rasters.jl
```diff
@@ -0,0 +1,114 @@
...
+    colsize!(fig.layout, 1, Aspect(1, aspect))
+    colgap!(fig.layout, 10)
+    return fig
+end
+
+function rplot(ras::Raster{T,3,<:Tuple{<:SpatialDim,<:SpatialDim,D}};
+    X=X, Y=Y,
+    xlabel = "lon",
+    ylabel = "lat",
+    clabel = "z",
```
asinghvi17, line 67: Can we pick these up from the band names?

asinghvi17 (commented):
Looked through this briefly and had a couple of comments/questions.  I'll try to add the rplot stuff in a weak dep which is also loaded by Requires in my PR.

.github/workflows/Documenter.yml
lazarusA: This for docs, later it will be used for plotting spheres. I was planning in also adding recipes for that.

src/plot_rasters.jl
lazarusA: This is how the main themes in Makie are defined isn't?

src/plot_rasters.jl
lazarusA: Yes, that's how the recipe in Plots works, I think.

src/plot_rasters.jl
lazarusA: And plan is to have a nice theme for Rasters, this the first iteration 😉

src/plot_rasters.jl
rafaqz: Maybe best not to use `lon` and `lat`, as its confusing if the units are something else. If we can verify the crs is in lat/lon then its fine

.github/workflows/Documenter.yml
asinghvi17: Smallish (720x360) spheres should be fine with CairoMakie ;)

src/plot_rasters.jl
asinghvi17: No, generally they are also defined as functions (to avoid someone accidentally changing the default attributes, and to avoid global variables).

Would you like to add this to MakieThemes instead of here, so we can start collecting all the themes of Makie in that repo?

lazarusA: Oh, maybe? I mean do we want to add another dependency? Maybe not. Transfer the theme there should be OK, but having solo here should be fine, and the convention, I think.

rafaqz:
I rabased this against master if anyone want to pick it up again:
https://github.com/rafaqz/Rasters.jl/tree/lazarusA-new_docs

PR #389, opened 2023-02-15
Add Makie recipes

asinghvi17:
This PR adds Makie recipes (converts) to Rasters.  I've tested the 2D recipes extensively, but not the 1D recipe since I don't have a sample handy.

TODOs:
- [x] Tests
- [x] Plotting `Raster{<: ColorTypes.Colorant, 2} as an image
- [ ] Plotting 3D rasters as volumes
- [x] Potentially integrating other recipes which I have overlooked


rafaqz:
Probably should talk to @lazarusA about this, it's already implemented in his PR as far as I know.

asinghvi17:
Sounds good.  I think this is pretty useful as a baseline anyway, so should work well.

We could put your recipe in a Requires block for now?  That seems like the simplest solution which still allows users on older versions of Julia to access it (unlike glue packages which you can only use if on 1.9+)

lazarusA:
Requires. Yes, that sounds great. Please, carry on 😁

rafaqz:
Great. @asinghvi17 in julia 1.9 we can have a weak dependency on Makie.jl with Requires.jl as a fallback for older versions. So it's fine to just write it up with Requires.jl for now.

asinghvi17:
Just implemented the extension package...fairly straightforward.  Here's what I have so far:

```julia
fig, ax1, plt1 = rplot(some_2d_raster)
ax2, plt2 = rplot(fig[1, 2], some_other_2d_raster; draw_colorbar = true, Axis = (; aspect = DataAspect()))
fig
```


(forgive the bad colorbar, am running with some other Makie changes also)

lazarusA:
I used `rplot` but I wanna to simply use `plot`, BUT, is there a problem with the `plot` function defined by Makie? could we extend that one? or?

asinghvi17:
I'd rather not extend `plot`, since we're not sure how the interface to create "Figure recipes" (as opposed to the plot recipes we have now) will evolve.  This also makes it easier to deprecate `rplot` at some point in the future, in favour of a Makie plot.

lazarusA:
Yeah, I figured. Thanks.

rafaqz:
Yeah figure recipes is the reason I never finished writing makie plotting recipes, theyre the main thing I actually use day in my work.

But `rplot` is pretty obscure.

asinghvi17:
I guess we could technically inject code here

https://github.com/MakieOrg/Makie.jl/blob/406b48d31f37efad17b8671fdc4d1ad0b78c9399/src/figureplotting.jl#L31

but it's a bit annoying since you don't have access to any converts and that approach seems kind of fragile.  Might make sense to leave the extension code in a PR, and merge the rest for now?

lazarusA:
I personally don't like this approach of leaving things behind and come back later, months could pass or even years and even the simplest thing, even if not ideal is never used. I prefer to have something working and then when a new thing, a better thing is available do the switch.

rafaqz:
Yeah let's just do it. Maybe don't export `rplot` for now while we work it out, so it's clear that it's not the long term syntax.

asinghvi17:
Sounds good, will do that then.  I believe the only things left are:
1. Figure out a way to do this without creating an internal gridlayout, so that axes can actually align properly
2. Implement rplot for 3d rasters
3. Add some tests from #380

lazarusA:
I went for Gridlayout because I wanted to have also the Colorbar for each heatmap. Which kinda works.

asinghvi17:
Oh definitely.  You can theoretically do e.g. `Colorbar(fig[i, j, Right()], ...)` for example, which adds to the protrusion of the axis so that other axes align to its inside, and not the outside.  I tried that, but wasn't able to pad the colorbar so that there is some whitespace between it and the axis.  Will try it again tonight.

In re your issue: yeah, I figured it was something like this.  Being able to add the colorbar as a protrusion should help.

lazarusA:
Ohh, yes putting the Colorbar in the protrusions could be cleaner. Definitely!

tomchor:
> This PR adds Makie recipes (converts) to Rasters. I've tested the 2D recipes extensively, but not the 1D recipe since I don't have a sample handy.

I'd like to help out in this PR if possible, so let me know if you still need to test 1D or other types data. I'm new to Rasters.jl and relatively new to Makie.jl, but I'd like to take this opportunity to learn a bit (and hopefully eventually also implement Makie extensions in DimensionalData.jl).

asinghvi17:
Thanks!  I definitely appreciate that...I do need to test 1D data and generally have someone take a look at the plotting experience and what could be improved, since my use case is basically restricted to 2D rasters.  I'm also not using v1.9, so am not sure if this works smoothly there or not.

BTW: @lazarusa, I figured out how to get colorbars in protrusions working :D - turns out that you just have to set the alignmode to `Outside(left::Real, right::Real, bottom::Real, top::Real)`.  I'll implement that and we can have a proper heatmap grid for 3d rasters.

asinghvi17:
Success!
iTerm2 Jo9hTh

I still need to figure out how we should handle rasters with missing here (convert to NaN?) but this should otherwise be good to go.

lazarusA:
like in [here](https://github.com/lazarusA/Rasters.jl/blob/c76e8c50676daf1fc7e1e754eea2fa67bcdd15ec/src/plot_rasters.jl#L74)
```
ras = replace_missing(ras, missingval=NaN)
```

lazarusA:
and, also, in this example we do have access to the axis right? We need to be able to `hidedecorations` afterwards if we want. In this case `X` in  fig[1,2] and `Y` in `fig[1,2]` and `fig[1,1]`. As the defaults is ok as it is, I think, but being able to hide them would be nice.

rafaqz:
Makie won't plot missing values?

I guess

asinghvi17:
ah, the axes should also be linked, I guess.  Will have to do that.

Since I've typed these for <: Real, missing values won't be accepted here.  Should probably have another convert for rasters with Missing, which can convert the type intelligently.

rafaqz:
In the Plots.jl recipes only the bottom left plot has labels and ticks.

asinghvi17:
Do we want that?  Since we have full control over the layout, we could do a facet-like thing where all plots at the bottom have x decorations, and all plots at the left have y decorations.

lazarusA:
I like the way it is now. We could hide them afterwards. I actually think that the default in plot is not a 100 % OK. I would rather have the defaults from Makie now.

rafaqz:
That style of Plots.jl tiles is copied from another raster package (I forget witch lol), so its just a matter of taste. For me the current layout wastes too much space, its nearly half white padding.

This is a default from terra:
image

And rasters

image

They skip the middle ticks - so the facet plot idea.


In Plots.jl it wass hard to control the spacing enough on all backends so just removing them was the best option. This plots in stars has them all but very tight with no waste of space, maybe we can do that as another option:

image

Repeating the colorbar is unnecessary if they are all slices of the same raster. We should also make sure the color limits are synchronised in that case. For a stack we do need to repeat the color bar.

lazarusA:
Thanks @asinghvi17, this is looking already too great!!

> In Plots.jl it wass hard to control the spacing enough on all backends

Indeed, here hopefully things will be easier. 
I did some checks, and the following examples are not working with this PR.

> Repeating the colorbar is unnecessary if they are all slices of the same raster. We should also make sure the color limits are synchronised in that case. For a stack we do need to repeat the color bar.

Agree. 
> One colorbar if things are for the same raster. 
However, in this case maybe we will be better off with a Colorbar in a normal figure position, and not in the protrusion as is now.
> individual colorbars for a stack [which is the current behaviour, although now is not for a stack]


## 1d
```Julia
using Downloads: download
url = "https://www.unidata.ucar.edu/software/netcdf/examples/tos_O1_2001-2002.nc";
filename = download(url, "tos_O1_2001-2002.nc");
A = Raster(filename)
lines(replace_missing(A[Y(Near(20.0)), Ti(1)], missingval=NaN)) # fails
```

## RasterStack
I think there is not a recipe for this case yet.

```Julia
using Rasters: Center
st = read(RasterStack(WorldClim{Climate}; month=1))
ag = aggregate(Center(), st, (Y(20), X(20)); skipmissingval=true, progress=false)
Rasters.rplot(ag) # fails!!!
```

### missingval = NaN issues
```Julia
evenness = Raster(EarthEnv{HabitatHeterogeneity}, :evenness)
rnge = Raster(EarthEnv{HabitatHeterogeneity}, :range)

# Roughly cut out South America
sa_bounds = X(-88 .. -32), Y(-57 .. 13)
sa_evenness = evenness[sa_bounds...]
# Extend range to match the whole-world raster
sa_range = extend(sa_evenness; to=rnge)
Rasters.rplot(replace_missing(sa_range, missingval=NaN)) # fails, but it works with missingval=0
```

```Julia
A = Raster(WorldClim{Climate}, :prec; month=1)
a = Rasters.rplot(replace_missing(A, missingval=NaN)) # fails, with 0 works

flags = Dict(
    :tr => [2.0, 2.0],
    :r => :near,
)
b = Rasters.rplot(replace_missing(warp(A, flags), missingval=NaN)) # fails, with 0 works
```

tomchor:
Ah, it seems like @lazarusA beat me in testing the 1D cases. It can confirm his example fails for me using Julia 1.8. You can find the [error log here](https://pastebin.com/wqRMnDN7). I also turned his snippet into a [self-contained MWE](https://pastebin.com/GqwQZAZs) if that makes it easier for anyone.

asinghvi17:
OK, so the way I see it there are two different potential inputs: Raster, whose colormap should be the same throughout the plots, and RasterStack, which should have different colormaps.  Users should also be able to specify where the colormap should be in both cases.

I guess some of the issues are because even after `replacemissing!`, the raster is still typed as `Union{FloatXY, Missing}` which causes the dispatch to fail.  Maybe we should run `replacemissing!` on rasters as a general rule.  Is there a sufficiently speedy dispatch for this on rasters which have data guaranteed to be of a nonmissing type?

rafaqz:
`replace_missing` without the bang is what you want? It should remove the union type.

 We also can't alter raster data for plotting.

lazarusA:
> Is it possible to specify multiple colormaps and colour limits for a stack?
I was thing about this last week, and maybe defining a custom colormap cycle for stacks should be the simplest solution.

rafaqz:
Haha that merged quote got me so confused.

A cycle sounds good. You mean by default different colormaps for each plot?

lazarusA:
> You mean by default different colormaps for each plot?

Yes, either way the user should change them later, but it would be nice to have some by default even if don't apply to all cases.

asinghvi17:
I wouldn't want to cycle automatically, rather the user do that instead.  We can allow a `colormap` parameter which is either a vector of anything which `to_colormap` works on (symbol, string, Vector{RGBAf}, cgrad(...)) or a single such object.  These could be distributed per plot.

asinghvi17:
download-7
download-8
download-6

The examples @lazarusa gave should work now (a lot of issues were in `replace_missing`, but the style clearly leaves a lot to be desired).

I introduced 3 new kwargs: `colormap`, `colorrange`, and `link_colorranges`.  You can set `colormap` to any vector of colormaps and it should do the job as above.


I just need to figure out the theming a bit better.  Here's the last example with smaller colorbar ticklabels:

download-9

lazarusA:
For colorbars I would do 

```
tickalign = 1
width = ticksize = `same value`
```
but, this should be part of `theming`, i.e. the Raster's theme.

As for the aspect ratios is a matter of figure size, which could be adjusted on the fly, so that things fit properly. And that's it, things should be pretty ok.

It's off, didn't colorbars share the hight with the main plot?

rafaqz:
The colorbars look much better with the small font. 

I think stylistically removing most of the X/Y will be the biggest improvement now.

The axis tick fonts could also be smaller.

asinghvi17:
@lazarusA: I thought so too, but I guess it doesn't work that well for DataAspect axes.  I can see the use of a theme now, since the optimal style here seems to deviate a bit from Makie's default style.

@rafaqz: yes, will try to do some facet-like tick label removal next, as well as giving the x ticklabels some rotation.

lazarusA:
totally agree. All these should be part of `theme_rasters`.

rafaqz:
this is looking so good!!

asinghvi17:
@rafaqz do you know why some of the colorranges have really low extrema?  See e.g. the 1e-38 values in some of the color scales.  Besides this, I think that once the theme is implemented this should be ready to go.

rafaqz:
That's usually the sign of a wrong missing value - especially 1e-38 is commonly used as one. 

Maybe you need to run `replace_missing` earlier?

asinghvi17:
Hmm, this is present even before `aggregate`
```julia
julia> st = read(RasterStack(WorldClim{Climate}; month=1))
RasterStack with dimensions: 
  X Projected{Float64} LinRange{Float64}(-180.0, 179.833, 2160) ForwardOrdered Regular Intervals crs: WellKnownText,
  Y Projected{Float64} LinRange{Float64}(89.8333, -90.0, 1080) ReverseOrdered Regular Intervals crs: WellKnownText,
  Band Categorical{Int64} 1:1 ForwardOrdered
and 7 layers:
  :tmin Float32 dims: X, Y, Band (2160×1080×1)
  :tmax Float32 dims: X, Y, Band (2160×1080×1)
  :tavg Float32 dims: X, Y, Band (2160×1080×1)
  :prec Int16 dims: X, Y, Band (2160×1080×1)
  :srad UInt16 dims: X, Y, Band (2160×1080×1)
  :wind Float32 dims: X, Y, Band (2160×1080×1)
  :vapr Float32 dims: X, Y, Band (2160×1080×1)


julia> st.tmin
2160×1080×1 Raster{Float32,3} tmin with dimensions: 
  X Projected{Float64} LinRange{Float64}(-180.0, 179.833, 2160) ForwardOrdered Regular Intervals crs: WellKnownText,
  Y Projected{Float64} LinRange{Float64}(89.8333, -90.0, 1080) ReverseOrdered Regular Intervals crs: WellKnownText,
  Band Categorical{Int64} 1:1 ForwardOrdered
extent: Extent(X = (-180.0, 179.99999999999997), Y = (-90.0, 90.0), Band = (1, 1))
missingval: -3.4e38
crs: GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]
values: [:, :, 1]
           89.8333  89.6667  89.5     89.3333  89.1667  89.0     88.8333  88.6667  88.5     88.3333  …  -88.5     -88.6667  -88.8333  -89.0     -89.1667  -89.3333  -89.5     -89.6667  -89.8333  -90.0
 -180.0    -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38     -23.2122  -22.962   -21.5438  -21.3953  -22.7165  -22.5142  -22.3545  -22.1597  -19.9805  -20.338
 -179.833  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38     -24.023   -24.8357  -22.2917  -22.13    -23.496   -23.2797  -23.11    -22.905   -21.058   -21.01
    ⋮                                           ⋮                                            ⋮       ⋱              ⋮                                                 ⋮                           
  179.667  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  …  -25.55    -26.4258  -23.7405  -23.5748  -25.0317  -24.807   -24.6325  -24.4398  -22.4585  -22.41
  179.833  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38  -3.4f38     -23.93    -23.7178  -22.241   -22.0825  -23.462   -23.2607  -23.1095  -22.9375  -20.7255  -21.0588
```

not sure what the potential solution could be here.

rafaqz:
The `missingval` is displayed as `Float64` (that's a bug n `show` I just found) but its actually `Float32` `-3.4f38`, so the same as the missing values in the array?

They are not automatically replaced with `missing`, because for a lot of things its 10x slower. We should make a default that is at some point - but getting the design right is a little tricky.

rafaqz:
Yeah it would more accurately be called `replace_missingval`

asinghvi17:
Looks correct now!  Works flawlessly with themes too...

download

rafaqz:
Amazing!

For tests we can just run the functions, I don't think we need to actually verify the output is correct.

And is it possible to squeeze that color bar to the same height as the plot axis? It could also be a bit thinner to save space, like R package do.

But also I can play with any of these things later its already really good as-is.

asinghvi17:
Yeah, I haven't actually added tests :D just some docs which should run.  Might increase the runtime of the docs build by a couple of minutes, though.

About the colorbar height: I tried to do this the obvious way and ended up getting a stack overflow, not totally sure how to fix that.  Might be worth looking at it later or maybe over MakieCon.

We can definitely put a colorbar width in the theme.  `Rasters.rplot(stack; Colorbar = (; width = 10), Axis = (; aspect = DataAspect(),))` will also allow you to adjust the colorbar width.

rafaqz:
I'll be at MakieCon! we totally can totally hack at this then. So I'm happy to merge as-is for now.

These docs are going to take an hour to build, it's already pretty heavy just with Plots.jl. But I'm fine with that, as long as it runs somewhere in the pipeline. 

It might also be good to protect the rest of the test suite from precompiling Makie.

rafaqz:
There is a bug with the Shapefile.jl test currently, I think its something to do with calling extent and not specializing on `Point` which has no MBR. I'll have to fix that to get this working unfortunately.

Project.toml
```diff
@@ -65,4 +67,11 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "CFTime", "DataFrames", "Plots", "SafeTestsets", "Shapefile", "Statistics", "Test"]
+test = ["Aqua", "CFTime", "CairoMakie", "DataFrames", "Plots", "Makie", "SafeTestsets", "Shapefile", "Statistics", "Test"]
```
rafaqz, line 70:
```suggestion
test = ["Aqua", "CFTime", "DataFrames", "Plots", "Makie", "SafeTestsets", "Shapefile", "Statistics", "Test"]
```

ext/RastersMakie.jl
```diff
@@ -0,0 +1,276 @@
...
+                        to_value(colorbar.layoutobservables.computedbbox[].widths[2])
+                    else
+                        to_value(colorbar.layoutobservables.computedbbox[].widths[1])
+                    end
+                else
+                    to_value(colorbar_padding)
+                end
+
+            colorbar.alignmode[] = if colorbar_position == Makie.Top()
+                Makie.Mixed(Makie.GridLayoutBase.RectSides{Union{Nothing, Float32, Makie.GridLayoutBase.Protrusion}}(nothing, nothing, colorbar_padding, 0f0))
```
rafaqz, line 146: Seems like a lot of duplication of this same line, you could define

`T = Makie.GridLayoutBase.RectSides{Union{Nothing, Float32, Makie.GridLayoutBase.Protrusion}}`

Project.toml
```diff
@@ -65,4 +67,11 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "CFTime", "DataFrames", "Plots", "SafeTestsets", "Shapefile", "Statistics", "Test"]
+test = ["Aqua", "CFTime", "DataFrames", "Plots", "Makie", "SafeTestsets", "Shapefile", "Statistics", "Test"]
```
rafaqz, line 70:
```suggestion
test = ["Aqua", "CFTime", "DataFrames", "Plots", "SafeTestsets", "Shapefile", "Statistics", "Test"]
```

rafaqz:
Codecov takes a bit of a hit not testing...

asinghvi17:
I guess we could pass a raster through the pipeline just to be sure...

ext/RastersMakie.jl
```diff
@@ -0,0 +1,276 @@
...
+- `colorbar_padding = Makie.automatic`: The amound of padding between the colorbar and its axis.  If `automatic`, then this is set to the width of the colorbar.
+- `title = Makie.automatic`: The titles of each plot.  If `automatic`, these are set to the name of the band.
+- `xlabel = Makie.automatic`: The x-label for the axis.  If `automatic`, set to the dimension name of the X-dimension of the raster.
+- `ylabel = Makie.automatic`: The y-label for the axis.  If `automatic`, set to the dimension name of the Y-dimension of the raster.
+- `colorbarlabel = ""`: Usually nothing, but here if you need it.  Sets the label on the colorbar.
+- `colormap = nothing`: The colormap for the heatmap.  This can be set to a vector of colormaps (symbols, strings, `cgrad`s) if plotting a 3D raster or RasterStack.
+- `colorrange = Makie.automatic`: The colormap for the heatmap.  This can be set to a vector of `(low, high)` if plotting a 3D raster or RasterStack.
+- `nancolor = :transparent`: The color which `NaN` values should take.  Default to transparent.
```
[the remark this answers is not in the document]
asinghvi17: Everything except colorbarlabel is - that one is really `cb = Colorbar(...); cb.label`

lazarusA:
where is the emoji for `merge` `merge` `merge`. We need one 😄 .  This is awesome!

rafaqz:
Can you rebase against main? The docs build should be fixed.

rafaqz:
Seems like some of the docs examples are breaking in the doctests.

lazarusA:
Let's not invest to much time on this. I want to set up the new docs, once this is merged.

rafaqz:
They're the only real tests of any of this so they need to work

asinghvi17:
Huh...looks like a doctest error in the doc build.

rafaqz:
We should probably just make that `extract` doctsest just a `julia` block, it's not reliable.

There is a CI error too

src/plotrecipes.jl
```diff
@@ -248,3 +248,134 @@ function DD.refdims_title(refdim::Band; issingle=false)
...
+    return (xs, ys, zs)
+end
+
+function MakieCore.convert_arguments(::MakieCore.DiscreteSurface, raw_raster::AbstractRaster{<: Union{Missing, Real}, 2})
+    raster = replace_missing(missing_or_float32.(raw_raster), missingval = NaN32)
+    ds = DD._fwdorderdims(raster)
+    A = permutedims(raster, ds)
+    x, y = dims(A)
+    xs, ys, zs = DD._withaxes(x, y, (A))
+    return (Makie.edges(xs), Makie.edges(ys), zs)
```
rafaqz, line 295: @asinghvi17 isn't this the problem? `Makie.edges` is called here and this method is called in the tests where we only have MakieCore

asinghvi17: Oops - guess I was testing and somehow pulled that in.  Have reimplemented the function.

rafaqz:
@asinghvi17 is this good to merge now? Would be good to let @lazarusA finish the docs PR

asinghvi17:
Yep, it should be good to go!

rafaqz:
Merged! 

Thanks so much, that was a huge effort.

asinghvi17:
Thanks!  Glad I could help :D

PR #390, opened 2023-02-16
New version with compat for ArchGDAL v0.10

asinghvi17:
Would be nice to avoid version conflicts (you can add master, but it's easier for new users this way)


rafaqz:
These doctest problems are unrelated.

PR #397, opened 2023-03-06
Breaking: write and better read of series

rafaqz:
This PR improves read of series and adds `write` to write mulitiple files.

- [x] write series of stacks - in nested folders ?
- [x] test

Breaking change here is that `force=true` has to be used to overwrite files. Once we are writing many files to nested directories it feels pretty bad just overwriting them all by default, so now `force` is required.


rafaqz:
bug is unrelated will fix elsewhere

PR #432, opened 2023-04-23
Fast point rasterize

rafaqz:
@felixcremer this should get your Rasters.jl datashading working a bit faster

```julia
using Rasters
using Extents
ps = [(rand(Float32) * 100, rand(Float32) * 100) for _ in 1:1e8]
@time rast = rasterize(count, ps; to=Extent(X=(0, 100), Y=(0, 100)), size=256);
  0.552263 seconds (18 allocations: 513.422 KiB)
```

Half a second for a hundred million points ;)

But it doesn't scale as well the [sorted algorithm](https://gist.github.com/rafaqz/a1706c16418bf875cc3aafc631dda49a) when you zoom in:

```julia
julia> @time rast = rasterize(count, ps; to=Extent(X=(0, 1), Y=(0, 1)), size=256);
  0.106193 seconds (18 allocations: 513.422 KiB)
```

So making lots of these tiles that cant be cached at higher zooms is still kinda slow. But I think its better to have the fastest algorithm for a single rasterization here as its the main use case.


felixcremer:
That works really nicely. With 5e5 points scaling is very good, with 5e6 it starts to get laggy and with 5e7 it is totally unusable, but  we would have to be smarter in caching the rasterisation result for the raster shading and I agree, that it would be best to have the fastest algorithm for single rasterisation in Rasters.jl. And we could think about using PyramidScheme on this first rasterisation  to get all of the other zoom levels from a first rasterisation.

PR #447, opened 2023-05-13
add slope, focal and cost distance with Stencils.jl

### LINKED ISSUE #410: Accumulation Cost or Cost distance feature
Thanks for this very valuable package. It is not a bug report but more of a feature question. Would it be feasible, or is there any interest in a **cost distance analysis** solution such as:
- _Accumulated Cost_ from [SAGA](https://saga-gis.sourceforge.io/saga_tool_doc/2.2.6/grid_analysis_0.html)
- _Cost distance_ from [R Terra](https://rspatial.github.io/terra/reference/costDist.html)
Thanks a lot

rafaqz:
Use Stencils.jl for fast parallel/GPU focal and slope filters, and a (single threaded but still fast) cost_distance model.

This PR will likely take a bit to be finalised and may need some changes to Stencils.jl.

Closes #410

### DIFF
--- a/ext/RastersStencilsExt/RaststersStencilsExt.jl
+++ b/ext/RastersStencilsExt/RaststersStencilsExt.jl
@@ -1,0 +1,7 @@
+using Stencils
+using Stencils: Window
+using Unitful
+
+include("cost_distance.jl")
+include("focal.jl")
+include("slope_aspect.jl")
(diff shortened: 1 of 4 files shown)


rafaqz:
After discussion with @evetion, this could instead go in:

https://github.com/Deltares/GeoArrayOps.jl/blob/master/src/spread.jl

And we could make it all Rasters.jl compatible, if it isn't already

PR #450, opened 2023-05-20
Breaking: Split up dependencies with package extensions

rafaqz:
This PR make ArchGDAL, NCDatasets, CoordinateTransformations, and HDF5 weak dependencies.

It gets `@time using Rasters` down under a second.

The rationale is:

- Very few people use rotated rasters, so should not have to pay for loading CoordinateTransformations (and StaticArrays)
- Often workflows using netcdf and gdal  are mutually exclusive
- The SMAP HDF5 loading is extremely niche, and good to get out of everyones way.
- RasterDataSources and the web download/unzip stack are also not used in lots of workflows, there is no need to pay the load time

The downside is now we need to do:

```julia
using Rasters, ArchGDAL
```
or 
```julia
using Rasters, NCDatasets
```

etc.

@tcarion, again this is going to mash your PR. But maybe not so much? the shared backend can probably live in `sources` if CommonDataModel.jl is light enough for extensions not to be worthwhile. Then the extensions for NCDatasets.jl and GRIBDatasets.jl will likely be pretty small.


ext/RastersArchGDALExt/RastersArchGDALExt.jl
```diff
@@ -0,0 +1,32 @@
+module RastersArchGDALExt
+
+import ArchGDAL,
+    DiskArrays,
+    Extents,
+    Missings
+
+using DimensionalData,
```
felixcremer, line 8: Shouldn't these using statements only load the module?
using DimensionalData: DimensionalData as DD

Then you would not need the const lines below.

rafaqz: I'm not sure that's actually shorter

felixcremer:
I think, that is a good idea. We should make sure, that people get a meaningful hint which package to load, when the Raster call fails.

rafaqz:
Yeah I have error messages saying what to load. Tgere should beca `ExtensionMissingErrir` in base to standardise it tho

rafaqz:
Thats great, lets keep it as a dep

PR #460, opened 2023-06-06
fix warp memory driver compression error

maxfreu:
`warp` didn't work because the memory driver in use does not support the compress option. I applied an easy fix.

### DIFF
--- a/ext/RastersArchGDALExt/gdal_source.jl
+++ b/ext/RastersArchGDALExt/gdal_source.jl
@@ -401,7 +401,7 @@ function _gdal_process_options(driver::AbstractString, options::Dict;
 )
     gdaldriver = AG.getdriver(driver)
     # set default compression
-    if !("COMPRESS" in keys(options)) && AG.validate(gdaldriver, ["COMPRESS=ZSTD"])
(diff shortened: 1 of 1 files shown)


maxfreu:
Yes, true. I'll see what I can do.

maxfreu:
Ok, that was quicker than expected. Maybe because the crop and mask functionality worked so nicely and fast ;-)

So: I added a test for warp and removed once instance of tests that were included twice.

test/runtests.jl
```diff
@@ -10,11 +10,7 @@ if VERSION >= v"1.9.0"
     # Aqua.test_project_toml_formatting(Rasters)
 end
 
-# Only test SMAP locally for now, also RasterDataSources because CI dowloads keep breaking
```
rafaqz, line 13: We do need to leave this code in ;)

rafaqz (approved):
Nice tests! Thanks for persevering with that.

test/runtests.jl
maxfreu: I didn’t look into the files, but it appears a second time further down, right? Should I still put it back?

rafaqz: Oh right its duplicated, no worries then

PR #465, opened 2023-06-09
Updated error message when attempting to read a file without the appropriate backend.

JoshuaBillson:
The previous error message for reading rasters without an imported backend was confusing. The new message informs callers of the change and suggests the appropriate backend to use.

The error now reads, "Rasters.jl requires backends to be loaded externally as of version 0.8. Run `import $packagename` to read $filename"

### DIFF
--- a/src/sources/sources.jl
+++ b/src/sources/sources.jl
@@ -59,5 +59,5 @@ function _open(f, filename::AbstractString; source=_sourcetype(filename), kw...)
 end
 function _open(f, T::Type, filename::AbstractString; kw...)
     packagename = SOURCE2PACAKGENAME[T]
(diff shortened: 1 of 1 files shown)


rafaqz:
I just realised that this means the errors are now different for different backends. There are many of these errors...

JoshuaBillson:
> I just realised that this means the errors are now different for different backends. There are many of these errors...
> 
> It would be better to write a generic error handler that did this.

I'm not sure I understand the problem. Don't we want the error messages to be different for each backend?

rafaqz:
No sorry I wasnt clear. 

There are other errors for ArchGDAL not related to file loading that are now now different... like for `resample`, and probably for CoordinateTransformations.jl when you load rotated files with GDAL

PR #472, opened 2023-07-05
fix typo in series filename swap

maxfreu:


rafaqz:
We should test this too

maxfreu:
Yes, but...

I definitely won't be able to do this before end of July. Would be cool to have it merged before that; maybe having an issue for "testing raster series with duplicate_first=true" is an option?

rafaqz:
Ok can you make the issue at least?

maxfreu:
I created the issue. Had to create another one first :cry: :laughing: Thanks for your work!

Issue #476, opened 2023-07-13
Wrong input for resample gives confusing error ArchGDALBackendException


felixcremer:
I am trying to resample an extent to a different CRS so that I don't have to resample all data points and I get this confusing error where it complainst that ArchGDAL should be loaded.
ArchGDAL is loaded as visible from [3] and [2]

I would prefer to have a proper MethodError which would give a hint on what other methods are available.

```julia
julia> resample(EPSG(4326),lookup(rasfull,:X))
ERROR: `Rasters.jl` requires backends to be loaded externally as of version 0.8. Run `import ArchGDAL` to fix this error.
Stacktrace:
 [1] resample(args::EPSG; kw::Base.Pairs{Symbol, Mapped{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, DimensionalData.Dimensions.LookupArrays.ForwardOrdered, DimensionalData.Dimensions.LookupArrays.Regular{Float64}, DimensionalData.Dimensions.LookupArrays.Points, DimensionalData.Dimensions.LookupArrays.Metadata{Rasters.NCDsource, Dict{String, Any}}, ProjString, EPSG, X{Colon}}, Tuple{Symbol}, NamedTuple{(:res,), Tuple{Mapped{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, DimensionalData.Dimensions.LookupArrays.ForwardOrdered, DimensionalData.Dimensions.LookupArrays.Regular{Float64}, DimensionalData.Dimensions.LookupArrays.Points, DimensionalData.Dimensions.LookupArrays.Metadata{Rasters.NCDsource, Dict{String, Any}}, ProjString, EPSG, X{Colon}}}}})
   @ Rasters ~/Documents/rqa_cscale/dev/Rasters/src/extensions.jl:4
[...]

PR #478, opened 2023-07-23
resample and warp now throw MethodError after loading ArchGDAL

JoshuaBillson:
As discussed in #476, `resample` and `warp` were throwing `BackendException` after loading `ArchGDAL` instead of `MethodError`. This is fixed by redefining the stubs when `ArchGDAL` is loaded.


rafaqz:
Closing in favour of #480

PR #480, opened 2023-07-25
Throw method error or backend exception correctly

rafaqz:
Builds on #478


src/extensions.jl
```diff
@@ -2,17 +2,25 @@
...
-        throw(BackendException("ArchGDAL"))
+    @static if isdefined(Base, :get_extension) # julia > 1.9
+        if isnothing(Base.get_extension(Rasters, :RastersArchGDALExt))
+            throw(BackendException("ArchGDAL"))
+        else
+            throw(MethodError(resample, args))
+        end
     else
-        throw(MethodError(resample, args))
+        throw(BackendException("ArchGDAL"))
```
felixcremer, line 12: Shouldn't this be throwing a MethodError, because on 1.8 ArchGDAL will be loaded anyways as a hard dependency?

PR #483, opened 2023-07-27
Catch error in write to remove broken file

felixcremer:
When the write command errors it still sometimes construct a broken file on disk.
This change should make sure, that  the write function would tidy up these broken files after faliure, because they are in the way for later analysis.

Do you think, that this is enough to capture all broken files?
Should this also be tested for other backends?

Locally I get some spurious failures which have nothing to do with the changes but are coming from some download time out errors.


rafaqz:
I think we need to first check the file didn't already exist! 

One error could be that `force=true` was not used and the file existed.

rafaqz:
Closing as stale, feel free to continue

PR #486, opened 2023-08-03
enable all virtual file systems

maxfreu:
Hi! I wanted to use gdal's /vsitar virtual file system and found only checks for curl and mem, so I added the rest. I also modified the if condition a bit to use `startswith`. I haven't added any test; the best I could do is to provide a tar file and check with that. But I won't be able to check s3, az etc as I don't have any data there.


rafaqz:
Great! maybe put all the virtual file system strings in a `const` above the function to keep things clean?

I'm ok with not testing  them, its not really feasable and we test vsicurl already.

maxfreu:
Done!

rafaqz:
Thanks!

maxfreu:
You're welcome :)

PR #487, opened 2023-08-04
docs finished

lazarusA:
this is it, it should be enough 😄 . Merge , merge 😆 . I believe all original topics are covered. Otherwise we should address any missing section in a different PR.


rafaqz:
Merge!!! MERGE!!!

lazarusA:
it's not clear to me why is failing? some data is not being downloaded ?

rafaqz:
Markdown not defined?

https://github.com/rafaqz/Rasters.jl/actions/runs/5765889329/job/15632750918#step:6:179

PR #494, opened 2023-08-12
test versioning

lazarusA:
this is just a test, let's see if it works.


evetion:
You'll probably don't see anything here, as the versions is setup to update on tags only. It's mike during the mkdocs build that makes sure the versions are there.

Best to test locally that the website has the version dropdown, and play round with `mike` locally (list, serve, maybe even deploy).

lazarusA:
yeah, but locally I don't see a way to see those versions with `mike`. What's your workflow (order of commands list) to test locally?

evetion:
Did you check https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/ and https://github.com/jimporter/mike? Being in the docs folder and checking out gh-pages is also a good way to debug. Deploying a dev build manually seems like the best test to do.

PR #503, opened 2023-08-27
Fixed skipmissing bug

JoshuaBillson:
There was a bug when calling `skipmissing` on a raster whose elements are floats and for which missingval = nothing.

The following code segment resulted in calling `isnan` on a `Nothing` value, which throws a `MethodError`.
```julia
function _missing(x::AbstractFloat, itr)
    if isnan(missingval(itr))
        return ismissing(x) || isnan(x)
    else
        return ismissing(x) || x == missingval(itr)
    end
end
```

### DIFF
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -35,7 +35,6 @@ end
 
 _missing(x, itr) = isequal(x, missingval(itr))
 _missing(x::Missing, itr) = true
-
 _missing(x::Nothing, itr) = false
 
 Base.IndexStyle(::Type{<:SkipMissingVal{T}}) where {T} = IndexStyle(T)
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -34,7 +34,6 @@ function Base.iterate(itr::SkipMissingVal, state...)
 end
 
 _missing(x, itr) = isequal(x, missingval(itr))
-
 _missing(x::Missing, itr) = true
 
 _missing(x::Nothing, itr) = false
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -38,9 +38,9 @@ function _missing(x::AbstractFloat, itr)
     if isnothing(missingval(itr))
         return false
     elseif isnan(missingval(itr))
-        return ismissing(x) || isnan(x)
+        return isnan(x)
     else
-        return ismissing(x) || x == missingval(itr)
+        return x == missingval(itr)
     end
 end
 
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -37,7 +37,7 @@ _missing(x, itr) = isequal(x, missingval(itr))
 
 _missing(x::Missing, itr) = true
 
-_missing(x::Nothing, itr) = true
+_missing(x::Nothing, itr) = false
 
 Base.IndexStyle(::Type{<:SkipMissingVal{T}}) where {T} = IndexStyle(T)
 Base.eachindex(itr::SkipMissingVal) =
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -33,21 +33,11 @@ function Base.iterate(itr::SkipMissingVal, state...)
     item, state
 end
 
-_missing(x, itr) = x == missingval(itr)
+_missing(x, itr) = isequal(x, missingval(itr))
 
 _missing(x::Missing, itr) = true
 
 _missing(x::Nothing, itr) = true
-
-function _missing(x::AbstractFloat, itr)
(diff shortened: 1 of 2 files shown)


rafaqz:
Thanks, yeah that code makes no sense, I cant see how `ismissing` is required either given the type signature.

rafaqz:
Can you write a test?

JoshuaBillson:
Sure, I'll add some test cases to the "skipmissing uses missingval" test set in Array.jl. I can also clean up the code by removing the `ismissing` calls, as I agree that they're redundant.

rafaqz:
Thanks, LGTM

src/skipmissing.jl
```diff
@@ -35,10 +35,12 @@ end
 
 _missing(x, itr) = ismissing(x) || x == missingval(itr)  # mind the order, as comparison with missing returns missing
```
rafaqz, line 36: I think we need the check on this method as well:
```suggestion
 function _missing(x, itr) 
    # mind the order, as comparison with missing returns missing
    if isnothing(missingval(itr))
        ismissing(x)
    else
        ismissing(x) || x == missingval(itr)
    end
end
```

rafaqz (commented):
I missed this yeaterday, we just need to make sure a `nothing` `missingval` works for all the methods of `_missing`

src/skipmissing.jl
JoshuaBillson: I think it should be fine as it is since `ismissing(nothing)` returns `false`, so there's no need for the first clause.

```julia
julia> ismissing(nothing)
false
```

Unless you want `skipmissing` to keep `nothing` values when `missingval=nothing`?

rafaqz: Hmm hard to know whats correct there. 

Myvidea was that `missingval` being `nothing` means there is no `missingval` behaviour at all, rather than the missing value is `nothing`.

JoshuaBillson:
I've added test cases for `Integer` arrays, which show that both `missing` and `nothing` are successfully removed by `skipmissing` when `missingval = nothing`.

I also simplified the logic by writing specialized `_missing` methods for `Nothing` and `Missing`:
```julia
_missing(x, itr) = isequal(x, missingval(itr))

_missing(x::Missing, itr) = true

_missing(x::Nothing, itr) = true
```
The result is that `missing` and `nothing` values will always be removed, `NaN` will be removed when `missingval = NaN`, and all other values will be removed when they equal `missingval` regardless of whether their types match.

src/skipmissing.jl
JoshuaBillson: It's a strange edge case. It seems there are three reasonable solutions:
1) Treat `nothing` like `missing` and remove it always.
2) Remove `nothing` only when `missingval = nothing`.
3) Never remove `nothing`.

rafaqz: If `isnothing(missingval)` we just don't set any `nodata` value in gdal:
 
https://github.com/rafaqz/Rasters.jl/blob/main/ext/RastersArchGDALExt/gdal_source.jl#L516-L531

It represents the case where there is no missing data in the raster.

So `nothing` being seen as a `missingval` could be inconsistent with this. Probably we should not remove `nothing` here either, if that also makes sense to you.

JoshuaBillson: Sure, I'll modify the code accordingly. If everything looks good to you after that, I think we can close this PR.

JoshuaBillson:
I've added test cases covering all `_missing` methods. `skipmissing` now implements the following logic:
1. `missing` is always removed.
2. `nothing` is never removed.
3. All other values, including `NaN`, are removed only when they're equal to `missingval`.

src/skipmissing.jl
```diff
@@ -33,14 +33,11 @@ function Base.iterate(itr::SkipMissingVal, state...)
...
-_missing(x, itr) = ismissing(x) || x == missingval(itr)  # mind the order, as comparison with missing returns missing
-function _missing(x::AbstractFloat, itr)
-    if isnan(missingval(itr))
-        return ismissing(x) || isnan(x)
-    else
-        return ismissing(x) || x == missingval(itr)
-    end
-end
+_missing(x, itr) = isequal(x, missingval(itr))
+
```
rafaqz, line 37:
```suggestion
```

maxfreu:
> Thanks, yeah that code makes no sense, I cant see how `ismissing` is required either given the type signature.

Oops, yeah that one was my fault, sorry.

rafaqz:
Oh sorry I didn't mean to be so harsh, I thought it was mine!

maxfreu:
Dont worry :)

PR #515, opened 2023-09-09
check dims in constructor

rafaqz:
DimensionalData.jl already has this, Rasters.jl should too


rafaqz:
Implemented elsewhere

PR #533, opened 2023-09-30
use BitArray for burning buffer

rafaqz:
@alex-s-gardner you could try this branch on your huge array without `threaded=false`

It should use 1/8th of the memory.


alex-s-gardner:
This branch speeds things up considerably but for the large use case threaded=false is still much faster

`Branch = Rasters v0.8.4`
threaded=true: 260.770344 seconds (1.24 M allocations: 336.461 GiB, 0.09% gc time, 6.75% compilation time)
threaded=false: 8.425360 seconds (876.21 k allocations: 10.768 GiB, 0.88% gc time)

`Branch = use_BitArray`
threaded=true: 30.647470 seconds (932.37 k allocations: 42.203 GiB, 0.98% gc time)
threaded=false: 2.602775 seconds (913.17 k allocations: 1.491 GiB, 2.14% gc time)

rafaqz:
Ok thats way lower allocation at least.

Maybe the threads are running into each other somehow, there is a lock in there but mostly it can be avoided. I'll have to look into that.

PR #539, opened 2023-10-19
add possibility to specify crop dimensions

maxfreu:
This adds the possibility to specify crop dimensions. Hope it looks ok :)

### DIFF
--- a/src/methods/crop_extend.jl
+++ b/src/methods/crop_extend.jl
@@ -92,7 +92,10 @@ function _crop_to(x, to; kw...)
         return _crop_to(x, ext; kw...)
     end
 end
-_crop_to(A, to::RasterStackOrArray; kw...) = _crop_to(A, dims(to); kw...)
+function _crop_to(A, to::RasterStackOrArray; kw...)
+    target_dims = get(kw, :dims, name.(dims(to)))
+    return _crop_to(A, dims(to, target_dims); kw...)
+end
 function _crop_to(x, to::DimTuple; kw...)
     # We can only crop to sampled dims (e.g. not categorical dims like Band)
     sampled = reduce(to; init=()) do acc, d
@@ -100,7 +103,7 @@ function _crop_to(x, to::DimTuple; kw...)
     end
     return _crop_to(x, Extents.extent(to); kw...)
 end
-function _crop_to(x, to::Extents.Extent; touches=false)
+function _crop_to(x, to::Extents.Extent; touches=false, kw...)
     ds = dims(x, map(key2dim, keys(to)))
     # Take a view over the bounds
     _without_mapped_crs(x) do x1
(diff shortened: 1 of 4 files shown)


src/methods/crop_extend.jl
```diff
@@ -92,15 +92,18 @@ function _crop_to(x, to; kw...)
         return _crop_to(x, ext; kw...)
     end
 end
-_crop_to(A, to::RasterStackOrArray; kw...) = _crop_to(A, dims(to); kw...)
+function _crop_to(A, to::RasterStackOrArray; kw...)
+    target_dims = get(kw, :dims, name.(dims(to)))
```
rafaqz, line 96: You can just do this in the keywords directly without `get`, just add a `dims` keyword.

rafaqz (changes requested):
Lools good

src/methods/crop_extend.jl
```diff
@@ -92,15 +92,18 @@ function _crop_to(x, to; kw...)
         return _crop_to(x, ext; kw...)
     end
 end
-_crop_to(A, to::RasterStackOrArray; kw...) = _crop_to(A, dims(to); kw...)
+function _crop_to(A, to::RasterStackOrArray; kw...)
```
rafaqz, line 95:
```suggestion
function _crop_to(A, to::RasterStackOrArray; dims=dims(to), kw...)
```

src/methods/crop_extend.jl
```diff
@@ -92,15 +92,18 @@ function _crop_to(x, to; kw...)
...
+    target_dims = get(kw, :dims, name.(dims(to)))
+    return _crop_to(A, dims(to, target_dims); kw...)
```
rafaqz, line 97:
```suggestion
    return _crop_to(A, DD.dims(to, dims); kw...)
```

src/methods/crop_extend.jl
rafaqz, line 96:
```suggestion
```

maxfreu:
This was a clear case of not seeing the forest for the trees :D I'll fix it!

maxfreu:
I noticed that there are unused variables in the crop functions, so it's likely that they don't work as you intended:

https://github.com/rafaqz/Rasters.jl/blob/88af714ec6194c321c6d1188e2c38c5e318318cf/src/methods/crop_extend.jl#L98-L100

https://github.com/rafaqz/Rasters.jl/blob/88af714ec6194c321c6d1188e2c38c5e318318cf/src/methods/crop_extend.jl#L104

rafaqz:
Good cqtch, the second one `ds` is just no longer needed, thats in DD now.

In the first `sampled` should be used instead of `to`. It will mostly just stop bands or other `Categorical` lookups being used to crop.

maxfreu:
Actually, cropping forward ordered categorical int64 band dimensions worked. But I changed it nevertheless, as there probably will be problems with string values.

rafaqz:
Yeah the idea was you basically never want to use crop for bands or categorical things, so passing in bands in the `to` template should be ignored

Issue #562, opened 2023-11-13
Projection of a tiff


noahrhodes:
I am loading a tiff file, but it doesn't have lat lon coordinates so correlate it with other data.  It seems to use a projection that I am unfamiliar with:
`"PROJCS[\"Sphere_ARC_INFO_Lambert_Azimuthal_Equal_Area\",GEOGCS[\"GCS_Sphere_ARC_INFO\",DATUM[\"D_Sphere_ARC_INFO\",SPHEROID[\"Sphere_ARC_INFO\",6370997,0]],PRIMEM[\"Greenwich\",0],UNIT[\"degree\",0.0174532925199433,AUTHORITY[\"EPSG\",\"9122\"]]],PROJECTION[\"Lambert_Azimuthal_Equal_Area\"],PARAMETER[\"latitude_of_center\",45],PARAMETER[\"longitude_of_center\",-100],PARAMETER[\"false_easting\",0],PARAMETER[\"false_northing\",0],UNIT[\"metre\",1],AXIS[\"Easting\",EAST],AXIS[\"Northing\",NORTH]]"`

The axes of the plot are around (-2e6,2.5e6),(-2.1e6,1e6), and I don't know what those units are based on. 
![plot_6](https://github.com/rafaqz/Rasters.jl/assets/15333359/998293c4-c3bb-414d-bc04-aaebdbfee05f)

I want to project it to use lat lon, I assume that `EPSG(4326)` might do the trick, but I get the error below. 

I have a couple of questions.
1. Is this is issue with the input tiff?  If so, how can I set the correct data projection for the input file when I read it?
2. Is this an issue with the reprojection? Should I be using a different format?

## Example code and error
```julia
using Rasters, ArchGDAL, GeoJSON, GeoInterface, Extents
wfpi = Raster("WFPI_data/emodis-wfpi_data_20191009_20191009.tiff")


using Makie, CairoMakie, GeoMakie
fig = Figure()
plot(fig[1, 1], wfpi)
fig  # figure doesn't have lat/lon indices

reproject(EPSG(4326),wfpi) # try to project the data

julia> reproject(EPSG(4326),wfpi)
ERROR: ArgumentError: Cannot reproject from: 
[...]

PR #563, opened 2023-11-15
document, test and add keyword method for `reproject`

rafaqz:
`reproject` and `resample` were both confusingly similar, but with different syntax.

They also didn't refer to each other in docs or errors.

This PR fixes that.

See #562, @noahrhodes hopefully this makes the distinction clearer and directs you when to switch from `reproject` to `resample`


felixcremer (commented):
I mainly looked at the docstrings.

PR #565, opened 2023-11-24
Better extract

rafaqz:
This PR mostly adds new keywords to `extract`

@tiemvanderdeure feel like reviewing this thing? 

I made it return a vector - this isn't breaking as its still an iterable and we never promised the exact type or iterable.

I took the opportunity to polish the (pretty basic) implementation a little. As well as the new keywords, `extract` now properly handles flattening multiple polygons/feature collections/nested vectors of geoms so you always get a single "table" return value, and handles empty or all-missing inputs. The only known bug is mixed points and geometries in one vector or will fail. This is noted in the doc, and its kinda rare and not worth the effort right now. 

Also once I squash the commits together you wont just be on that one empty commit but the whole thing, its just to get your name on the list.

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -38,7 +38,7 @@ obs = GBIF2.occurrence_search("Burramys parvus"; limit=5, year="2009")
 
 # use `extract` to get values for all layers at each observation point.
 # We `collect` to get a `Vector` from the lazy iterator.
-collect(extract(st, pnts; skipmissing=true))
+extract(st, obs; skipmissing=true)
 
 # output
 5-element Vector{NamedTuple{(:geometry, :bio1, :bio3, :bio5, :bio7, :bio12)}}:
@@ -56,7 +56,6 @@ function extract end
 function extract(x::RasterStackOrArray, data;
     dims=DD.dims(x, DEFAULT_POINT_ORDER), names=_names(x), geometry=true, index=false, kw...
 )
-    
     T = if geometry 
         keys = index ? (:geometry, :index, names...,) : (:geometry, names...,)
         NamedTuple{keys}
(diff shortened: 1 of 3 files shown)


rafaqz:
@visr have you seen this anywhere else? we are getting failures in GDAL.jl `gdalsetprojection` only on macosx:

```
 Got exception outside of a @test
  GDALError (CE_Failure, code 1):
  	Invalid TOWGS84 node
  ```

See here: https://github.com/rafaqz/Rasters.jl/actions/runs/6697436007/job/18560020921#step:5:949

**Comment by visr** [comment]
Odd that this only happens on MacOS. I haven't seen this error before.

File: src/methods/extract.jl [hunk 1]
```diff
@@ -32,12 +36,9 @@ st = RasterStack(WorldClim{BioClim}, (1, 3, 5, 7, 12)) |> replace_missing
 # Download some occurrence data
 obs = GBIF2.occurrence_search("Burramys parvus"; limit=5, year="2009")
 
-# Convert observations to points
-pnts = collect((o.decimalLongitude, o.decimalLatitude) for o in obs if !ismissing(o.decimalLongitude))
-
 # use `extract` to get values for all layers at each observation point.
 # We `collect` to get a `Vector` from the lazy iterator.
-collect(extract(st, pnts))
+collect(extract(st, pnts; skipmissing=true))
```
tiemvanderdeure, line 41: No point in using collect on a Vector, right?
```suggestion
# use `extract` to get values for all layers at each observation point.
(extract(st, pnts; skipmissing=true))
```

tiemvanderdeure:
Handling missings seems to work wonderfully when giving `missing`, but behaves unpredictably with `NamedTuples` containing `missing`.

E.g.
```julia
using Rasters
myrast = Raster(rand(X(10), Y(10)))
extract(myrast, ((X = missing , Y = missing), (X= 1, Y = 10))) # work but return Vector{Any}
extract(myrast, ((X= 1, Y = 10), (X = missing , Y = missing))) # errors
extract(myrast, ((X = missing , Y = 10),)) # Gives a StackOverflowError!
```

Or is this also what you mean by mixed geometries? Any point data that comes from a table and has some missing coordinates will look like this.

Maybe a solution could be to use a function that also skips geoms that contain `missing` instead of `Base.skipmissing`? It should be reasonable to treat something like `(X = missing, Y = missing)` in the exact same way as `missing`

I'll have time to look at this more tomorrow.

rafaqz:
We should force the return type to be at least `Vector{NamedTuple{keys,Any}}`

We are not actually using `Base.skipmissing`, but `filter` doing pretty much what you suggest when`skipmissing=true` is passed in. Otherwise I think people can just filter them later as a dataframe however they like.

rafaqz:
Handling this isn't possible because your points with missing coordinates are not valid GeoInterface geometries:
```julia
extract(myrast, ((X = missing , Y = missing), (X= 1, Y = 10)))
```

And to throw a better error there we need to disallow arbitrary objects like you tuple of points, its limited to an `AbstractArray` now

src/methods/extract.jl
```diff
@@ -47,66 +48,130 @@ collect(extract(st, pnts))
...
-function _extract(A::RasterStackOrArray, ::Nothing, geoms; kw...)
-    geom1 = first(skipmissing(geoms))
-    if GI.isgeometry(geom1) || GI.isfeature(geom1) || GI.isfeaturecollection(geom1)
-        (_extract(A, g; kw...) for g in geoms)
+_extract(T, A::RasterStackOrArray, geom; kw...) = _extract(T, A, GI.geomtrait(geom), geom; kw...)
+_extract(T, A::RasterStackOrArray, ::Nothing, geom; kw...) = throw(ArgumentError("$geom is not a valid GeoInterface.jl geometry"))
+function _extract(T, A::RasterStackOrArray, ::Nothing, geoms::AbstractArray; names, skipmissing=false, kw...)
+    # Handle empty / all missing cases
+    (length(geoms) > 0 && any(!ismissing, geoms)) || return T[]
+    geom1 = first(Base.skipmissing(geoms)) # TODO: will fail if `geoms` is empty or all missing
```
tiemvanderdeure, line 78: As is `extract(araster, [(valid_geom, invalid_geom)])` and `extract(araster, [(invalid_geom, valid_geom)])` return different errors. Explicitly checking for invalid geometries would solve this.

```suggestion
    # Handle cases with some invalid geometries
    invalid_geom_idx = findfirst(g -> !ismissing(g) && GI.geomtrait(g) === nothing, geoms)
    invalid_geom_idx === nothing || throw(ArgumentError("$(geoms[invalid_geom_idx]) is not a valid GeoInterface.jl geometry"))

    geom1 = first(Base.skipmissing(geoms))
```

test/methods.jl
```diff
@@ -256,35 +256,134 @@ end
...
+        @test extract(rast_m, p; skipmissing=true, index=true) == [
+            (geometry = (9.0, 0.1), index = CartesianIndex(1, 1), test = 1)
+            (geometry = (10.0, 0.1), index = CartesianIndex(2, 1), test = 3)
+        ]                                                         
+        # Empty geoms
+        @test extract(rast, []) == NamedTuple{(:geometry, :test),Tuple{Missing,Missing}}[]
+        @test extract(rast, []; geometry=false) == NamedTuple{(:test,),Tuple{Missing}}[]
+        # Missing coord errors
+        @test_throws ArgumentError extract(rast, [(0.0, missing), (9.0, 0.1), (9.0, 0.2), (10.0, 0.3)])
+        @test_throws ArgumentError extract(rast, [(X=0.0, Y=missing), (9.0, 0.1), (9.0, 0.2), (10.0, 0.3)])
```
tiemvanderdeure, line 358: Add another test if an invalid geometry is not the first geometry

```suggestion
        @test_throws ArgumentError extract(rast, [(0.0, missing), (9.0, 0.1), (9.0, 0.2), (10.0, 0.3)])
        @test_throws ArgumentError extract(rast, [(9.0, 0.1), (0.0, missing), (9.0, 0.2), (10.0, 0.3)])
        @test_throws ArgumentError extract(rast, [(X=0.0, Y=missing), (9.0, 0.1), (9.0, 0.2), (10.0, 0.3)])
```

PR #566, opened 2023-11-27
Bump patch version to 0.9.1

felixcremer:
This would enable to use NCDatasets 0.13 from the latest Rasters release.


rafaqz:
Gdal is broken on mac

Maybe we should restrict the version? I dont have a mac so I cant fix it.

PR #581, opened 2023-12-21
Fixed `extract` bug for `RasterStack`

JoshuaBillson:
Attempting to call `extract` on a `RasterStack` gives the following error:

```julia
ERROR: UndefVarError: `t` not defined
```

The issue seems to be in methods/extract.jl:131

```julia
@inline _prop_nt(st::AbstractRasterStack, I, names::NamedTuple{K}) where K = t[I][K]
```

Which I fixed with the following:

```julia
@inline _prop_nt(st::AbstractRasterStack, I, names::NamedTuple{K}) where K = st[I][K]
```


rafaqz:
Thanks! Guess we are missing some tests too

PR #592, opened 2024-01-19
fix common data model chunk propagation, source type propagation, and other things

rafaqz:
This PR fixes a bunch of minor issues with the CommonDataModel.jl conversion.

- chunking is not propagated out through CommonDataModel.jl so we hack around it by fetching the inner chunks
- source type was not propagated through the shared `FileArray` object so online files without extensions were reverting to gdal (but was not tested because we had reliability problems in the tests)
- some minor renaming of funcitons that still had the `_ncd` prefix in the commondatamodel.jl file.

I'll need to write a for more tests to lock this down.


ext/RastersGRIBDatasetsExt/gribdatasets_source.jl
```diff
@@ -14,3 +12,7 @@ function RA._open(f, ::Type{GRIBsource}, filename::AbstractString; write=false,
     ds = GRIBDatasets.GRIBDataset(filename)
     RA._open(f, GRIBsource, ds; kw...)
 end
+
```
rafaqz, line 15: @tcarion @Alexander-Barth this is the hack that is currently required to make disk arrays chunking propagate out from the internal `Varable` or `DataValues` objects.

Note that with your current implementation this means anyone using `CFVariable` can not do chunked reads of large datasets. The whole thing is read at once.

src/sources/commondatamodel.jl
```diff
@@ -31,24 +31,38 @@ const CDM_STANDARD_NAME_MAP = Dict(
...
 
-# Methods from CommonDataModel
+# We have to dig down to find the chunks as they are not immplemented
+# properly in the source packages, but they are in their internal objects.
+# We just make it work here even though it doesn't work in NCDatasets.jl or GRIBDatasets.jl
+DiskArrays.eachchunk(var::CFVariable) = _get_eachchunk(var.var.var)
+DiskArrays.haschunks(var::CFVariable) = _get_haschunks(var.var.var)
+
+function _get_eachchunk end
+function _get_haschunks end
```
rafaqz, line 63: @tcarion @Alexander-Barth this is where the hack is called to dig into the inner array that knows the actual chunk pattern. With the current design of NCDatasets.jl/GRIBDatasets.jl/CommonDataModel.jl anyone wanting chunking to work with CFVariable will need to do a hack like this.

(Note this is the `Rasters.CFVariable`, not type piracy)

PR #595, opened 2024-01-31
fix cellsize

tiemvanderdeure:
The cellsize function broke recently after a breaking change that caused `crs` to always return `nothing` for `Tuple`s.

This wasn't noticed because cellsize wasn't actually added to runtests.jl.

Maybe we want to consider also defining `crs` such that `crs(myraster)` and `crs(dims(myraster))` return the same thing. Unfortunately `dims(myraster) == NTuple` returns `false` so I'm not sure how to define in a way that is generic and specific to all Tuples of `Dimension`.

### DIFF
--- a/ext/RastersArchGDALExt/cellsize.jl
+++ b/ext/RastersArchGDALExt/cellsize.jl
@@ -48,9 +48,8 @@ function _area_from_coords(transform::ArchGDAL.CoordTransform, ::GI.LinearRingTr
     return _area_from_rads(GI.LinearRing(points))
 end
 
-function cellsize(dims::DimTuple)
-    xbnds, ybnds = DimensionalData.intervalbounds(dims)
-    # take CRS from Y if it has crs, else from X
+function cellsize(dims::Tuple{<:X, <:Y})
+    xbnds, ybnds = DD.intervalbounds(dims)
     dcrs = crs(dims)
     if convert(CoordSys, dcrs) == CoordSys("Earth Projection 1, 104") # check if need to reproject
         areas = [_area_from_coords(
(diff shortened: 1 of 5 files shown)


ext/RastersArchGDALExt/cellsize.jl
```diff
@@ -50,7 +50,8 @@ end
 
 function cellsize(dims::Tuple{X, Y})
```
rafaqz, line 51: Does this dispatch even work? that seems pretty limited

tiemvanderdeure: If you call cellsize on a raster it just takes the X and Y dimension out. But ideally this would be defined for any Tuple of Dimensions, I think. I'm just not sure how to dispatch on that?

rafaqz: But shouldn't it be `Tuple{<:X,<:Y}` at least? they have a type parameter

rafaqz: You can just use `DimTuple` for dispatch. But then you need to choose dims by name with `DD.dims(dims, X)`

tiemvanderdeure: I now changed this to `Tuple{<:X,<:Y}` and added a dispatch for `DimTuple` separately.

tiemvanderdeure:
I'm ont sure why the Documenter fails?

rafaqz:
Its an unrelated makie problem, we can ignore ut

tiemvanderdeure:
Good to go then?

PR #598, opened 2024-02-02
add TILED=YES creation option for gdal write

maxfreu:


maxfreu:
What I'm trying to do here is to detect whether 2D tiling was used in the source raster by checking if the blocks are divisible by 16 (which they have to be for GTiff). Traditional line-by-line compression does not fulfill this criterion, so no tiling is applied.

ext/RastersArchGDALExt/gdal_source.jl
```diff
@@ -406,7 +406,14 @@ function _process_options(driver::String, options::Dict; _block_template=nothing
     # creation options are driver dependent
 
     if !isnothing(_block_template) && DA.haschunks(_block_template) == DA.Chunked()
-        block_x, block_y = string.(DA.max_chunksize(DA.eachchunk(_block_template)))
+        block_x, block_y = DA.max_chunksize(DA.eachchunk(_block_template))
+        
+        if (block_x % 16 == 0) && (block_y % 16 == 0)
```
rafaqz, line 411: A comment explaining what this does would help

maxfreu:
Is there some keyword for avoiding CI?

rafaqz:
We can ignore the docs error. But this needs a test to merge.

maxfreu:
Yes, right. Tests will take some time, not before April I guess :(

rafaqz:
If been thinking about adding a `chunks` keyword to write that makes controling chunking easier, and standardised as much as possible accross netcdf and gdal backends.

`chunks` instead of `tiled` mostly because DiskArrays.jl calls them "chunks" already.

But this would allow your use case: `chunks=true` and also `chunks=(256, 256)` and generate the right keywords for all the backends where it can be done, and warn/error otherwise.

rafaqz:
I'm going to merge this as we need something similar but, probably I'll move the `TILED=YES` part to a keyword before release

maxfreu:
Cool, thanks :)

rafaqz:
Probably `chunks=(256, 256)` etc, but we could have `chunks=true` or `chunks=:detect` or similar to mean `TILED=YES`. Does that make sense?

maxfreu:
`chunks=(x,y)` makes sense. I'd opt for `chunks=true/false` to control automatic detection or no chunking. When I was new to julia I was traumatized by undocumented usage of symbols in Makie, so I'm still a bit hesitant at using them.

How would you deal with cases where the user gives chunks=something and possibly contradicting values in the creation options? Put a warning if `chunks` is given and the options dict contains `TILED` or blocksizes? Then one would have to select one of them.

PR #599, opened 2024-02-05
missingmask of all layers in rasterstack

tiemvanderdeure:
Changes the default behaviour of `missingmask` when applied on a RasterStack or RasterSeries, so the mask covers all layers.

It's probably possible to reduce the memory use of this operation by using missingmask!, but I can't really see how to guarantee that that works when not all dimensions are shared.

The dispatch on RasterSeries might run into errors if the dimensions are not shared between the layers, so maybe we don't want it?

### DIFF
--- a/src/methods/mask.jl
+++ b/src/methods/mask.jl
@@ -322,19 +322,6 @@ savefig("docs/build/missingmask_example.png"); nothing
 
 $EXPERIMENTAL
 """
-function missingmask_multilayer(layers::Union{<:AbstractRasterStack, <:AbstractRasterSeries}, to; kw...)
-    dest = _init_bools(to, Array{Union{Missing,Bool}}, layers; kw..., missingval=missing)
-    dest .= true
-    map(layers) do layer
-        missingval=_missingval_or_missing(layer)
-
-        broadcast_dims!(dest, dest, layer) do d, x
-            isequal(d, missing) || isequal(x, missingval) ? missing : true
-        end
-   end
-   return dest
-end
-
 function missingmask(stack::AbstractRasterStack; alllayers = true, to = dims(stack), kw...) 
     if alllayers
         missingmask_multilayer(stack, to; kw...)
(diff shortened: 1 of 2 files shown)


rafaqz:
Yeah it will have to be using `mask!`.

`broadcast_dims!` can probably handle this pretty easily. First make the mask array matching the size of the stack.
then map over it using `broadcast_dims!` on each layer to update the values.

Should only need one allocation.

tiemvanderdeure:
I definitely think we want missing values to propagate across dimensions. I can't come up with another consistent way to do this.

tiemvanderdeure:
Alright, tried my best. But it might be you can come up with some wizardry to make this even cleaner?

src/methods/mask.jl
```diff
@@ -321,8 +322,35 @@ savefig("docs/build/missingmask_example.png"); nothing
 
 $EXPERIMENTAL
 """
-missingmask(series::AbstractRasterSeries; kw...) = missingmask(first(series); kw...)
-missingmask(stack::AbstractRasterStack; kw...) = missingmask(first(stack); kw...)
+function missingmask_multilayer(layers::Union{<:AbstractRasterStack, <:AbstractRasterSeries}, to; kw...)
```
rafaqz, line 325: Why `_multilayer` ?

best to control this with a keyword

rafaqz:
Is there a test for mixed size stacks?

tiemvanderdeure:
I added some.

Right now `missingmask` will fail on a `RasterSeries` if some layers have dimensions that the first layer does not have. For a `RasterStack` it just calls `dims`, but we cannot do this for a `RasterSeries`. 

What do you think the most logical behaviour would be in that case?

rafaqz:
RasterSeries is for fixed dimensions so thats fine.

tiemvanderdeure:
In that case I think we're good to go.

PR #601, opened 2024-02-15
boolmask for rasterstack/series

tiemvanderdeure:
This implements boolmask methods for rasterstacks and rasterseries, with a alllayers keyword that defaults to `true`.

It is the exact same behaviour as already exists for `missingmask`.


tiemvanderdeure:
Good to go now I think!

PR #606, opened 2024-03-09
Bump NCDatasets to 0.14 and CommonDataModel to 0.3

felixcremer:
This bumps NCDatasets to 0.14.
I had to add the ignore to stale_deps check to make the tests run locally. 
This is only needed because we need Requires for the extensions in Julia 1.8 or lower.
This is currently failing because of two fails in the NCDataset tests, I can have a look at this on Monday.

Why are there no CompatHelpers for NCDAtaset, is it because it is a weakdep?


rafaqz:
Closed in favor of #607

Issue #611, opened 2024-03-12
missingval keyword for RasterStack


tiemvanderdeure:
It doesn't seem possible to pass a `missingval` keyword to `RasterStack` at the moment. Would this be hard to implement?

I ran into this with the CHELSA dataset, which has `UInt16` type but a missingval of `-99999`, which throws an error unless `missingval` is passed explicitly (not sure if this is a bug or not in itself?).

E.g.
```
using RasterDataSources, Rasters
RasterStack(CHELSA{BioClim}, (1,2); missingval = -99999.0)
```

This workaround does work, of course:
`rs = RasterStack(Raster.(CHELSA{BioClim}, (1,2); missingval = -99999.0)...)`

rafaqz:
How can there be any missing values if its below zero?

rafaqz:
Ok looking at it now its actually `Int16`, and its has the correct missingval of -32768...

But you are right, its not possible to change it to -9999. Will fix that

```julia
julia> st = RasterStack(CHELSA{BioClim}, (1, 2); missingval = -9999)
╭─────────────────────────╮
│ 43200×20880 RasterStack │
├─────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Projected{Float64} LinRange{Float64}(-180.00013888885002, 179.99152633785002, 43200) ForwardOrdered Regular Intervals{Start},
  → Y Projected{Float64} LinRange{Float64}(83.99152708185001, -90.00013888884999, 20880) ReverseOrdered Regular Intervals{Start}
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── layers ┤
  :bio1 eltype: Int16 dims: X, Y size: 43200×20880
  :bio2 eltype: Int16 dims: X, Y size: 43200×20880
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── raster ┤
  extent: Extent(X = (-180.00013888885002, 179.99985967115), Y = (-90.00013888884999, 83.99986041515001))
  missingval: (bio1 = -32768, bio2 = -32768)
  crs: GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.01745329251994
[...]

PR #613, opened 2024-03-19
add extra checks to _missingval_from_gdal when x is a float

tiemvanderdeure:
When a missingval is taken from a GDAL datasource and is a float, but the datatype is an integer, check that it is within the type bounds before attempting to convert.

Fixes an error mentioned in https://github.com/rafaqz/Rasters.jl/issues/611

I don't think we need to add a test for such a simple change?


rafaqz:
If theres no test we can delete it later with no consequences

PR #614, opened 2024-03-19
better filestack performance

### LINKED ISSUE #226: Polynomial RasterStack creation time
Hi! It seems like the RasterStack creation time is polynomial wrt the number of input files. Is this to be expected? If yes, why? In my special case I'm loading Sentinel 2 data from one tile of a [force data cube](https://force-eo.readthedocs.io/en/latest/howto/datacube.html) using following code:

<details>
  <summary>Code</summary>

```julia
using Rasters
using Glob
using Dates
using Gnuplot

l2_folder = "/mnt/nas/projects/klimba/force/l2/europe/"
tile_folder = l2_folder * "X0031_Y0024/"
files = readdir(glob"*BOA.tif", tile_folder)
filenames = basename.(files)
times = [f[1:8] for f in filenames]
# times = DateTime.([f[1:8] for f in filenames], "yyyymmdd")

fs = filesize.(files)/10^9
##
nmax = 70
loading_times = Float64[]
for n in 1:2:nmax
    println(round(sum(fs[1:n])), " GB")
    t = @elapsed RasterStack(files[1:n], keys=Ti(Symbol.(times[1:n])));
    push!(loading_times, t)
end

##
begin
@gp "set title 'RasterStack creation time (Rasters v0.1.0)'"
@gp :- "set y2tics"
@gp :- "set key top left; set ytics nomirror"
@gp :- "set xl 'Number of files'"
@gp :- "set yl 'Load time [s]'"
@gp :- "set y2l 'Cumulative file size [GB]'"
@gp :- 1:2:nmax loading_times "w l t 'time'"
@gp :- 1:2:nmax [round(sum(fs[1:n])) for n in 1:2:nmax] "w l axes x1y2 t 'size'"
save(term="pngcairo size 640,480", output="~/RasterStackCreationTime.png")
end
```  
</details>

In a fresh session it looks like this:

RasterStackCreationTime

To avoid the x-y problem: What I'm actually trying to do is to lazily (spatial and temporal) index into a force data cube, which is essentially a bunch of tifs in folders. How would you tackle the problem? Probably using RasterStacks is going the wrong way? But maybe discourse is a better place for that question.

rafaqz:
This PR overhauls both stack load and compile time. It uses type-unstable vectors for as long as possible at load to avoid compiling lots of functions on huge tuples.

The 130 layer stack in the tests now takes ~6 seconds to load the first time, and 0.5 seconds after that. `getindex` after using `open` is now essentially the same as using NCDatasets or GRIBDatasets directly.

We should encourage using `open(st) do st` blocks as much as possible for `lazy=true` stacks, its much faster now.

Closes #226

(Sorry this took 2 years @maxfreu :sweat_smile: )


maxfreu:
Thanks for improving this! But you shouldn't work so long on Sundays ;-)

rafaqz:
Not to worry, github tells me I have enjoyed *eleven* entire, healthy, commit-free Sundays in just the last year alone!

PR #621, opened 2024-04-03
fix reading a lazy rasterstack

tiemvanderdeure:
fixes bugs in _open_layers, so `read(::RasterStack)` no longer errors.

Also fixes `filename(::RasterStack)` so it returns a `NamedTuple` of file names, rather than throwing an error.

 I added some tests for both changes.

### DIFF
--- a/test/sources/gdal.jl
+++ b/test/sources/gdal.jl
@@ -548,7 +548,7 @@ end
         gdalstack_lazy = RasterStack((a=gdalpath, b=gdalpath); lazy=true)
         @test Rasters.isdisk(gdalstack_lazy.a)
         @test Rasters.isdisk(gdalstack_lazy.b)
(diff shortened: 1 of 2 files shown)


test/sources/gdal.jl
```diff
@@ -548,9 +548,14 @@ end
         gdalstack_lazy = RasterStack((a=gdalpath, b=gdalpath); lazy=true)
         @test Rasters.isdisk(gdalstack_lazy.a)
         @test Rasters.isdisk(gdalstack_lazy.b)
+        gdalstack_read = read(gdalstack_lazy)
+        read!(gdalstack_lazy, gdalstack_read)
```
rafaqz, line 552: It always gets me that `read` methods have flipped arguments because of IO arg order rules...

PR #622, opened 2024-04-03
extract over rows if rows are geometries

tiemvanderdeure:
adds a check to extract such that if `Tables.rows(x)` iterates geometries (e.g. X-Y points), the points are extracted. Previously this would throw an error.

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -64,7 +64,7 @@ function extract(x::RasterStackOrArray, data;
         NamedTuple{keys}
     end
     names = NamedTuple{names}(names)
-    if Tables.istable(data) && !(data isa AbstractVector{<:GeoInterface.NamedTuplePoint})
+    if !(data isa AbstractVector{<:GeoInterface.NamedTuplePoint}) && Tables.istable(data)
         geomcolnames = GI.geometrycolumns(data)
         if isnothing(geomcolnames) || !in(first(geomcolnames), Tables.columnnames(Tables.columns(data)))
             throw(ArgumentError("No `:geometry` column and `GeoInterface.geometrycolums(::$(typeof(data)))` does not define alternate columns"))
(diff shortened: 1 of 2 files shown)


src/methods/extract.jl
```diff
@@ -66,9 +66,17 @@ function extract(x::RasterStackOrArray, data;
     names = NamedTuple{names}(names)
     if Tables.istable(data)
         geomcolnames = GI.geometrycolumns(data)
-        isnothing(geomcolnames) && throw(ArgumentError("No `:geometry` column and `GeoInterface.geometrycolums(::$(typeof(data)))` does not define alternate columns"))
-        geometries = Tables.getcolumn(Tables.columns(data), first(geomcolnames))
-        _extract(T, x, geometries; dims, names, kw...)
+        if isnothing(geomcolnames)
```
rafaqz, line 69: Hmm is it ever `nothing` ? Or will it fall back to `:geometry` ? 

I think we may need to check `first(geomcolnames) in Tables.columnames(data)`

tiemvanderdeure:
I think I cleaned everything up now. To handle geometries/tables with geometries consistently throughout, we should probably add an internal function that takes `data` and `geometrycolumns` as arguments and:
- returns `data` if `data` is an iterable (or just `AbstractVector`) of geometries
- returns geometries if `geometrycolumns` is specified
- returns geometries if `data` has a column corresponding to `GeoInterface.geometrycolumns(data)` 
- errors with an informative error message otherwise.

src/methods/extract.jl
```diff
@@ -64,12 +64,14 @@ function extract(x::RasterStackOrArray, data;
         NamedTuple{keys}
     end
     names = NamedTuple{names}(names)
-    if Tables.istable(data)
+    if Tables.istable(data) && !(data isa AbstractVector{<:GeoInterface.NamedTuplePoint})
```
rafaqz, line 67:
```suggestion
    if !(data isa AbstractVector{<:GeoInterface.NamedTuplePoint}) && Tables.istable(data)
```
The cheap one goes first. In many cases compiler wont be able to prove that `istable` has no side effects so it will have to run it.

rafaqz:
Yeah that function is a good idea.

PR #627, opened 2024-04-07
Specify the units of `res` in docstrings

asinghvi17:
Not sure if this is accurate but it was a bit confusing to me...

### DIFF
--- a/src/methods/shared_docstrings.jl
+++ b/src/methods/shared_docstrings.jl
@@ -11,7 +11,7 @@ const SIZE_KEYWORD = """
     Only required when `to` is not used or is an `Extents.Extent`, and `res` is not used.
 """
 const RES_KEYWORD = """
-- `res`: the resolution of the dimensions, a `Real` or `Tuple{<:Real,<:Real}`.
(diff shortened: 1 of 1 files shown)


rafaqz:
"Or degrees" maybe?

asinghvi17:
Are the units just in terms of the units of X and Y?

src/methods/shared_docstrings.jl
```diff
@@ -11,7 +11,7 @@ const SIZE_KEYWORD = """
     Only required when `to` is not used or is an `Extents.Extent`, and `res` is not used.
 """
 const RES_KEYWORD = """
-- `res`: the resolution of the dimensions, a `Real` or `Tuple{<:Real,<:Real}`.
+- `res`: the resolution of the dimensions (usually in meters), a `Real` or `Tuple{<:Real,<:Real}`.
```
rafaqz, line 14:
```suggestion
- `res`: the resolution of the dimensions (often in meters or degrees), a `Real` or `Tuple{<:Real,<:Real}`.
```

PR #631, opened 2024-04-09
Fix typos

felixcremer:
Fix two typos.

### DIFF
--- a/ext/RastersCoordinateTransformationsExt/affineprojected.jl
+++ b/ext/RastersCoordinateTransformationsExt/affineprojected.jl
@@ -4,7 +4,7 @@ function AffineProjected(f;
     AffineProjected(f, data, metadata, crs, mappedcrs, paired_lookup, dim)
 end
 
-DD.dim(lookup::AffineProjected) = lookup.dim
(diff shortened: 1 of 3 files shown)


ext/RastersCoordinateTransformationsExt/affineprojected.jl
```diff
@@ -4,7 +4,7 @@ function AffineProjected(f;
     AffineProjected(f, data, metadata, crs, mappedcrs, paired_lookup, dim)
 end
 
-DD.dim(lookup::AffineProjected) = lookup.dim
+DD.dims(lookup::AffineProjected) = lookup.dim
```
rafaqz, line 7:
```suggestion
DD.dim(lookup::AffineProjected) = lookup.dim
```

This one isn't a typo! 

Maybe its silly to have but its only for lookups and at the `Lookup` module level `dims` is not available - and it really shouldn't be a Lookup module function. `dims` also returns a Tuple that we don't have for a lookup.

felixcremer: Yeah I got confused because when I load DimensionalData I didn't find DD.dim

PR #635, opened 2024-04-15
test allocations when writing

tiemvanderdeure:
Add tests for the number of allocations when writing. This should catch bugs as in https://github.com/rafaqz/Rasters.jl/pull/633, that do not throw an error but cause excessive allocations and very slow writing speeds.

### DIFF
--- a/test/sources/ncdatasets.jl
+++ b/test/sources/ncdatasets.jl
@@ -240,7 +240,8 @@ end
             geoA = read(ncarray)
             @test size(geoA) == size(ncarray)
             filename = tempname() * ".nc"
-            write(filename, geoA)
+            write(filename, geoA; force = true)
+            @test (@allocations write(filename, geoA; force = true)) < 1e4
             @testset "CF attributes" begin
                 @test NCDatasets.Dataset(filename)[:x].attrib["axis"] == "X"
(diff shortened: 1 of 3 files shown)


test/sources/ncdatasets.jl
```diff
@@ -479,6 +488,7 @@ end
         @test metadata(saved)["advection"] == "Lin & Rood"
         @test metadata(saved) == metadata(st) == metadata(ncstack)
         @test all(first(DimensionalData.layers(saved)) .== first(DimensionalData.layers(st)))
+        @test (@allocations write(filename, st)) < 1e6 # writing a rasterseries/stack has no force keyword
```
rafaqz, line 491: does `force` not work here?

rafaqz (approved):
GTM

test/sources/ncdatasets.jl
tiemvanderdeure: No, that's a bug.

rafaqz: Would be nice to have a `@test_broken` in the tests for with is the kw PR doesn't fix it. I have a bunch of tests writing stacks a second time with `force=true` and they work.

PR #638, opened 2024-04-18
Faster geom extract

### LINKED ISSUE #637: Faster polygon and line `extract`
Currently `extract` is very badly optimised for lines, and also pretty slow for polygons.

Rather than rasterizing the whole raster and reading it, it should just push values to a vector as it goes.

The actual rasterization doesn't even appear in the profile, so there is maybe a few order of magnitude available here.

rafaqz:
Closes #637

A few simple changes that make line extract 3 order of magnitude faster:

- use a view of the geometry extent before masking
- type stability

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -74,14 +74,16 @@ function extract(x::RasterStackOrArray, data;
     end
 end
 
-function _extract(T, A::RasterStackOrArray, geom::Missing; names, kw...)
+function _extract(::Type{T}, A::RasterStackOrArray, geom::Missing; names, kw...) where T
     [_maybe_add_fields(T, A, map(_ -> missing, names), missing, missing)]
 end
-_extract(T, A::RasterStackOrArray, geom; kw...) = 
+_extract(::Type{T}, A::RasterStackOrArray, geom; kw...) where T = 
     _extract(T, A, GI.geomtrait(geom), geom; kw...)
-_extract(T, A::RasterStackOrArray, ::Nothing, geom; kw...) =
+_extract(::Type{T}, A::RasterStackOrArray, ::Nothing, geom; kw...) where T =
     throw(ArgumentError("$geom is not a valid GeoInterface.jl geometry"))
-function _extract(T, A::RasterStackOrArray, ::Nothing, geoms::AbstractArray; names, skipmissing=false, kw...)
+function _extract(::Type{T}, A::RasterStackOrArray, ::Nothing, geoms::AbstractArray; 
+    names, skipmissing=false, kw...
+) where T
     # Handle empty / all missing cases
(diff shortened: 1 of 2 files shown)


src/methods/extract.jl
```diff
@@ -102,28 +104,37 @@ function _extract(T, A::RasterStackOrArray, ::Nothing, geoms::AbstractArray; nam
...
+) where T
     # Make a raster mask of the geometry
-    B = boolmask(geom; to=commondims(A, DEFAULT_POINT_ORDER), kw...)
+    template = view(A, Touches(GI.extent(geom)))
+    ods = otherdims(A, DEFAULT_POINT_ORDER)
+    if length(ods) > 0
+        template = view(template, map(d -> rebuild(d, firstindex(d)), ods)) 
+    end
+    B = boolmask(geom; to=template, kw...)
+    offset = CartesianIndex(map(first, parentindices(template)))
```
tiemvanderdeure, line 131: This should be `parentindices(parent(template))`. A `view` of a `Raster` is a `AbstractArray` and not a `SubArray` and therefore `parentindices(template)` doesn't work.

tiemvanderdeure: actually even `offset = CartesianIndex(map(first, parentindices(template))) - CartesianIndex(1,1)` because of the 1-based indexing

src/methods/extract.jl
```diff
@@ -102,28 +104,37 @@ function _extract(T, A::RasterStackOrArray, ::Nothing, geoms::AbstractArray; nam
...
     # Add a row for each pixel that is `true` in the mask
-    rows = (_maybe_add_fields(T, A, _prop_nt(A, I, names), I) for I in CartesianIndices(B) if B[I])
+    rows = (_maybe_add_fields(T, A, _prop_nt(A, I + offset, names), I) for I in CartesianIndices(B) if B[I])
```
tiemvanderdeure, line 133: the second `I` should also be `I + offset` I think?

tiemvanderdeure (changes requested):

src/methods/extract.jl
rafaqz: yep

src/methods/extract.jl
```diff
@@ -128,9 +128,9 @@ function _extract(::Type{T}, A::RasterStackOrArray, t::GI.AbstractGeometryTrait,
         template = view(template, map(d -> rebuild(d, firstindex(d)), ods)) 
     end
     B = boolmask(geom; to=template, kw...)
-    offset = CartesianIndex(map(first, parentindices(template)))
+    offset = CartesianIndex(map(first, parentindices(parent(template))))
```
tiemvanderdeure, line 131: this still needs `- CartesianIndex(1,1)`

PR #639, opened 2024-04-22
collapse missingval when constructing rasterstack if possible

tiemvanderdeure:
Before this PR, a call to `RasterStack` will always return a stack where `missingval` is a `NamedTuple`, which looks kind of clunky if all rasters in a big stack have the same `missingval`.

This just checks if all `missingval`s are the same, and sets the missingval to just a value if that's the case.


tiemvanderdeure:
This should all be covered by existing tests, so good to go as far as I'm concerned

PR #643, opened 2024-04-25
bump DimensionalData compat

alex-s-gardner:
compat bumped to 0.27


rafaqz:
If only it was that easy lol

alex-s-gardner:
Ya, I just tried... :-(

PR #656, opened 2024-05-03
precise step

tiemvanderdeure:
An attempt to avoid the worst floating point errors when calculating step.

@rafaqz let me know what you think about this heuristic

### DIFF
--- a/src/sources/commondatamodel.jl
+++ b/src/sources/commondatamodel.jl
@@ -390,11 +390,10 @@ function _cdmspan(index, order)
         # Calculate step, avoiding as many floating point errors as possible
         st = Base.step(Base.range(Float64(first(index)), Float64(last(index)); length = length(index)))
         st_rd = round(st, digits = Base.floor(Int,-log10(eps(eltype(index))))) # round to nearest digit within machine epsilon
-        st_rd ≈ st ? st_rd : st # keep the rounded number if it is very close to the original
+        isapprox(st_rd, st; atol = eps(eltype(index))) ? st_rd : st # keep the rounded number if it is very close to the original
     else
         index[2] - index[1]
     end
-    
     for i in 2:length(index)-1
         # If any step sizes don't match, its Irregular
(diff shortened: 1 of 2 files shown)


src/sources/commondatamodel.jl
```diff
@@ -385,12 +385,16 @@ end
...
     length(index) == 1 && return Regular(zero(eltype(index))), Points()
-    
-    # Calculate step, avoiding as many floating point errors as possible
-    st = Float64(Base.step(Base.range(first(index), last(index); length = length(index))))
-    st_rd = round(st, digits = Base.floor(Int,-log10(eps(eltype(index))))) # round to nearest digit within machine epsilon
-    step = st_rd ≈ st ? st_rd : st # keep the rounded number if it is very close to the original
 
+    step = if eltype(index) <: AbstractFloat
+        # Calculate step, avoiding as many floating point errors as possible
+        st = Float64(Base.step(Base.range(first(index), last(index); length = length(index))))
```
rafaqz, line 391:
```suggestion
        st = Base.step(Base.range(Float64(first(index)), Float64(last(index)); length = length(index)))
```

The `Float64` conversion needs to happen before the division in `range`, not after

rafaqz:
I want to merge this but we probably should have a test of the edge case where the rounding actually does something, just to demonstrate that we do this at all

rafaqz:
Something weird happening in the `resample` tests to cause a segmentation violation (signal 11).

https://github.com/rafaqz/Rasters.jl/actions/runs/9148702377/job/25152012187?pr=656#step:5:933

tiemvanderdeure:
Funily enough I can't reproduce this by just running tests/resample.jl, so not sure what is going on?

rafaqz:
Maybe its only segfaulting on a specific GDAL version. Welcome to the joy of C pointers

rafaqz:
@tiemvanderdeure if you can rebase this and push again GDAL will probably work now

PR #657, opened 2024-05-03
match to-dims with resample

tiemvanderdeure:
This PR ensures that the `x` and `to` have the same dimensions after resampling, if only `to` is provided.

Using size rather than resolution as a fallback means floating point error does not affect the outcome.

If only `to` is provided, the dimensions of the resampled raster will be set exactly match the dimensions of `to`.

### DIFF
--- a/ext/RastersArchGDALExt/resample.jl
+++ b/ext/RastersArchGDALExt/resample.jl
@@ -108,8 +108,3 @@ function resample(A::RasterStackOrArray;
 end
 
 _size_and_res_error() = throw(ArgumentError("Include only `size` or `res` keywords, not both"))
-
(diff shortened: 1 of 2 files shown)


rafaqz:
Maybe we are not dropping the band dimension properly, that should happen in `warp` but maybe isn't?

Usually we check if there was a band in the original raster and `dropdims(rast, Band())` if not.

tiemvanderdeure:
No the problem was setting the dims to `dims(to)` without checking if there were other dims than just X/Y

tiemvanderdeure:
Finally got tests to pass!

rafaqz (approved):

PR #659, opened 2024-05-07
add inverse keyword for mask

rafaqz:


src/methods/mask.jl
```diff
@@ -1,4 +1,9 @@
 
+const INVERSE_KEYWORD = """
+- `inverse`: invert the mask, so that areas no missing in `with` are
+    masked, and areas missing in `with` are masked.
```
asinghvi17, line 4: Do the other packages also use `inverse`?  I feel like this should be `invert`...

rafaqz: Me too! But raster/terra do.

Turns out its `invert` in rasterio so I'll switch it

asinghvi17: Maybe mention
```suggestion
const INVERSE_KEYWORD = """
- `inverse`: invert the mask, so that areas no missing in `with` are
    masked, and areas missing in `with` are masked.  Equivalent to 
    `inverse` in R's `raster` and `terra` packages.

File: ext/RastersArchGDALExt/resample.jl [hunk 2]
```diff
@@ -88,6 +88,10 @@ function resample(A::RasterStackOrArray;
     # resample with `warp`
     resampled = warp(A, flags; kw...)
 
+    if !isnothing(to)
```
tiemvanderdeure, line 91: This must have been added accidentally, right?

rafaqz: Haha yes maybe from when we were demoing that change

PR #665, opened 2024-05-10
check crs (and sampling) in cellsize and reproject

tiemvanderdeure:
Adds checks to `reproject` and `cellsize`, so both error if an object without a CRS is passed, and `cellsize` additionally errors if an object with Points sampling is passed.


tiemvanderdeure:
Very annoying that WorldClim makes tests fail here as well...

rafaqz:
Ah right. The docs is ok but we shouldn't have them in gdal tests

rafaqz:
Its fixed on main, can you rebase and push?

PR #674, opened 2024-05-26
make it work with latest DD and Makie

lazarusA:


lazarusA:
@rafaqz tests passed. Can we merge this one. I will update docs, afterwards, and then we could release a new version.

PR #675, opened 2024-05-26
La/new vite docs: ready to merge

lazarusA:
Let's see if CI is working.


lazarusA:
I don't know why the DD restriction. Locally, is running ok.

rafaqz:
I think the DD problem might be from GitHub actions caching? I had this somewhere else too.

asinghvi17:
If you have admin privileges on the repo, you can delete all caches in the "Actions" tab

asinghvi17:
Looks like UC Davis's server is down again

lazarusA:
good. That fixed the error. But we need to increase the timeout on Downloads. Locally also takes a while. Ideas?

felixcremer:
I added a ` Pkg.Registry.update();` call in the github action. This should solve these problems also in the future. 
See [JuliaDataCubes/YAXArrays.jl@`7474202` (#390)](https://github.com/JuliaDataCubes/YAXArrays.jl/pull/390/commits/74742026f04e200ad941812ebbf76e30c1553d25)

rafaqz:
Ok those are just worldclim/ucdavis outage problems now. 

(We should really purge WorldClim from everything it happens too much)

felixcremer:
The problem with caching is, that this would be redistribution which is allowed only after permission  https://worldclim.org/about.html. When we are using WorldClim as examples, we should find other example datasets that are more reliable or which we could host ourselves.

lazarusA:
@felixcremer @rafaqz any suggestions for a different dataset provider for those examples. So that I can fix this and finish the PR.

rafaqz:
Not sure. WorldClim files are relatively small which helps download time

rafaqz:
Chelsea is a direct replacement, we could use the lowest resolution?

lazarusA:
@rafaqz that should do it. Let's see if data is not longer an issue.

rafaqz:
Now there's a real error ;)

ERROR: LoadError: invalid '@docs' syntax: @docs data_sources

lazarusA:
Ok. Now is a path issue/missing images. The `doctest` creates images, that are not being saved at the right place or rather before they are needed by Vitepress. Will look into this later.

asinghvi17:
│ ERROR: You must set `ENV["RASTERDATASOURCES_PATH"]` to a path in your system

lazarusA:
it is, 
here: https://github.com/rafaqz/Rasters.jl/blob/484cb172e35ccace8bf9106ec0f2d21f381a097c/.github/workflows/Documenter.yml#L47 and 

(twice!)
here: 

https://github.com/rafaqz/Rasters.jl/blob/484cb172e35ccace8bf9106ec0f2d21f381a097c/docs/make.jl#L15

lazarusA:
the error is only for the `doctests`. Really confused now.

rafaqz:
Does the folder exist tho? It won't make one for you

lazarusA:
we r good now: https://rafaqz.github.io/Rasters.jl/previews/PR675/

PR #688, opened 2024-06-27
Bump RasterDatasource version

felixcremer:
I did not fully understand why there was a breaking release of RasterDataSources. Hopefully, this is just going through.


rafaqz:
We updated the default version of Chelsa or something, so you get different data

PR #695, opened 2024-07-12
Breaking: Standardise cf standards and missingval handling

rafaqz:
This PR standardises cf and missingval behaviour accross all sources.

New keywords are:
- `cf=true`: apply `offset` and `scale` if the  are not zero and one.
- `maskingval=missing` this is the value the missingval will be converted to if there is one.

I don't like either of these names! if you can think of anything better please suggest something.

@felixcremer @meggart this is the new diskarrays object that does 2-way cf and missingval/maskingval conversions

https://github.com/rafaqz/Rasters.jl/compare/cf?expand=1#diff-1478a88a8bdcffcf7f556cc1865dbe18d69b5c03f75b4481e1f5889e8a6b6b73


Rapsodia86:
How about inverting it? Instead of "cf" call it "raw". If raw=false, offset and scale are applied, if raw=true, offset and scale are NOT applied?

And raw=false would be the default?

rafaqz:
Currently I've got `scaled=true` as the default instead of `cf`. `raw` might suggest the missing value is also not transformed?

Main concern with `navalue` is it sounds like R, and isn't really used in julia

Rapsodia86:
Yes, you are right that with "raw" one may expect that data were not touched at all.

Indeed, "navalue" originates from my heavy R usage! And I agree NA is not very julia way. 

Point taken:)

rafaqz:
Thanks for the input though, good to have all the options to choose from

rafaqz:
Coming back to this I like how short and obvious `raw=true` is!

But it to mean no scaling and no masking - so it will override everything else.

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -0,0 +1,20 @@
+module GeometryOpsDimensionalDataExt
```
asinghvi17, line 1: Should we close the PR that does this in GeometryBasics?

src/create.jl
```diff
@@ -1,41 +1,344 @@
...
-function create(filename, T, A::AbstractRaster; 
-    name=name(A), metadata=metadata(A), missingval=missingval(A), kw...
+Create a new, uninitialised [`Raster`](@ref) or [`RasterStack`](@ref).
+
+If `filename` is a `String` it will be created on disk, and opened lazily.
+If it is `nothing` or not passed, an in-memory `Raster` will be created.
+
+If type is a `Type` return value is a `Raster`. The `eltype` will usually be `T`, except
+where `scale` and/or `offset` keywords are used or a `missingval` of a different type is specified, 
+in which case `T` will depend on the tyepe promotion of `scale`, `offset` and `missingval` with `T`.
```
asinghvi17, line 13:
```suggestion
in which case `T` will depend on the type promotion of `scale`, `offset` and `missingval` with `T`.
```
typo

src/create.jl
```diff
@@ -1,41 +1,344 @@
...
+`maskingval` will also affect the `eltype` of the openeded raster if you `create` to a file.
+
+If types is a `NamedTuple` of types, the result will be a `RasterStack`. In this case `fill` and 
+`missingval` can be single values (for all layers) or `NamedTuple` with the same names to specify per-layer.
+
+`f!` will be applied to the `Raster` or `RasterStack` while it is stil open after creation, 
+to avoid opening it twice. The return value of `f!` is disguarded but modifications
```
asinghvi17, line 20:
```suggestion
to avoid opening it twice. The return value of `f!` is disregarded but modifications

File: src/create.jl [hunk 4]
```diff
@@ -1,41 +1,344 @@
...
+to the `Raster` or the `RasterStack` layers will be written to disk or changd in memory.
```
asinghvi17, line 21:
```suggestion
to the `Raster` or the `RasterStack` layers will be written to disk or changed in memory.

File: src/create.jl [hunk 5]
```diff
@@ -1,41 +1,344 @@
+const TypeNamedTuple = NamedTuple{<:Any,<:Tuple{Vararg{Type}}}
 
+"""
+    create([f!], [filename], template; kw...)
```
asinghvi17, line 4: When should I use this as opposed to `Raster(data, dims; ...)`?  What are the advantages beyond efficiency in writing to file?

asinghvi17: Or is the direct constructor going to call `create`?

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
rafaqz: No I meant st have added that by mistake lol

src/create.jl
rafaqz: Well `Raster(filename)` reads an existing file... create creates one. 

So this is best for building something new from scratch, `Raster` is best for opening something, rewrapping a DimArray, etc

asinghvi17: Gotcha, I guess I'm not entirely clear on what to use when creating a purely in memory dataset.  Does the pure `Raster(data, dims; kwargs...)` constructor have the same options?  Should it?   That's the most intuitive option IMO to create an in memory raster...

src/methods/shared_docstrings.jl
```diff
@@ -125,3 +125,82 @@ const CHECKMEMORY_KEYWORD = """
...
+    This is common where data has been convert to e.g. UInt8 to save disk space.
+    To ignore `scale` and `offset` metadata, use `scaled=false`. If `scale` and
+    Note: `offset` are `1.0` and `0.0` they will be ignored and the original type will 
+    be used even when `scaled=true`. This is because these values may be fallback 
+    defaults and we do not want to convert every `Real` array to larger `Float64` values.
+"""
+
+const COERCE_KEYWORD = """
+- `coerce`: where `scale` and/or `offset` are present during `setindex!` to disk, 
+    coerce values to the disk type. `convert` is the default, but `round`, `trunc` or
```
asinghvi17, line 155: What's the disk type?

src/methods/shared_docstrings.jl
```diff
@@ -125,3 +125,82 @@ const CHECKMEMORY_KEYWORD = """
...
+    when you know the value is not specified or is incorrect. This will *not* change any
+    values in the raster, it simply assigns which value is treated as missing. 
+"""
+
+const MASKINGVAL_KEYWORD = """
+- `maskingval`: A value to convert `missingval` to, by default `missing`. If this is set it 
+    will be the return value of `missingval(raster)` - `maskingval` becomes the new `missingval`.
+    Setting `maskingval` to `nothing` means no masking will occur, and the original `missingval` 
+    will be the final `missingval`. This can give better performance than using `missing`. 
+    Another efficient option is to use e.g. `zero(eltype(raster))` to replace missing values with zero.
```
asinghvi17, line 170: I can see why you picked this but I'm not totally happy about the name.  Still, I can't see any other nice options.  `missingval` would have been nice but it's already in use.  I'd say we should go for it.  A sufficiently good docstring will clarify all problems :P 

Maybe emphasize that this does what `replace_missing` does?  
```suggestion
- `maskingval`: A value to convert `missingval` to, by default `missing`. If this is set it 
    will be the return value of `missingval(raster)` - `maskingval` becomes the new `missingval`.
    This is the same thing that `replace_missing(raster, some_value)` does.
    Setting `maskingval` to `nothing` means no masking will occur, and the original `missingval` 
    will be the final `missingval`. This can give better performance than using `missing`. 
    Another efficient option is to pass e.g. `zero(eltype(raster))` to replace missing values with zero.
```

src/methods/shared_docstrings.jl
rafaqz: Oh it should be eltype

src/methods/shared_docstrings.jl
rafaqz: It's also what CommonDataModel uses so there's precedent... But sharing the mask part with the `mask` function is kinda weird for me

rafaqz: I've decided to scrap `maskingval` and just have `missingval`, but allow `sourcemissingval => destmissingval` as a syntax

It's a bit more breaking but I think nicer long term

PR #698, opened 2024-07-15
add global checkmem

### LINKED ISSUE #690: Allow Users to Read Data Exceeding Free Memory
I've noticed that recent versions of `Rasters.jl` no longer allow me to read rasters into memory when their size exceeds the value reported by `Sys.free_memory()`. This is creating issues for me, because it appears that `free_memory` does not detect swap space, so it reports that I only have 200 MB available. While I can circumvent this to a certain extent by setting `lazy=true`, I cannot save the transformed rasters back to disk without throwing a `GDALError` with the message `Pointer 'hDS' is NULL in 'GDALGetRasterCount'`. There should be an option to override this behaviour, with the possibility of a warning message when reading data that exceeds the reported memory space.

**Edit:** On further examination, it appears that `Sys.free_memory()` is just incorrect. According to my activity monitor, I should have 4GB of memory available, but `Sys.free_memory()` is reporting 150MB. This is on an M1 MacBook Pro running Julia 1.10.4 on macOS Ventura.

rafaqz:
Closes #690 

@JoshuaBillson hopefullly this will fix your problem? Would be good if you can try out the branch when the tests are passing.

`checkmem=false` keyword to `Raster` or `read` will turn off the check locally.

`Rasters.checkmem!(false)` will turn it off globally.


JoshuaBillson:
Reading rasters now works both when passing `checkmem=false` and setting `checkmem!` globally.

rafaqz:
Thanks good to hear

PR #704, opened 2024-08-02
fix vector element type in zonal pre-allocation

DanTanAtAims:
If the aggregation function returns missing with the first geometry mask in `zonal` then the function will error with

```
nested task error: cannot convert a value to missing for assignment
``` 

In order to determine the element type of the vector for pre-allocation, `_alloc_zonal` calculates the first value of the vector to be returned separately and uses it's type when pre-allocating.
See [here](https://github.com/rafaqz/Rasters.jl/blob/main/src/methods/zonal.jl#L118-L123)
```julia
function _alloc_zonal(f, x, geom, n; kw...)
    z1 = _zonal(f, x, geom; kw...)
    zs = Vector{Union{Missing,typeof(z1)}}(undef, n)
    zs[1] = z1
    return zs
end
```
However, if the first element is missing then `zs = Vector{Union{Missing, Missing}}(undef, n)`. 

This proposed change uses the first non-missing type as the type of the Vector.

### DIFF
--- a/src/methods/zonal.jl
+++ b/src/methods/zonal.jl
@@ -108,16 +108,31 @@ function _zonal(f, x::RasterStackOrArray, ::Nothing, data; progress=true, thread
     geoms = _get_geometries(data, geometrycolumn)
     n = length(geoms)
     n == 0 && return []
-    zs = _alloc_zonal(f, x, first(geoms), n; kw...)
-    _run(1:n, threaded, progress, "Applying $f to each geometry...") do i
+    zs = _alloc_zonal(f, x, geoms, n; kw...)
+    start_index = findfirst(x -> !ismissing(x), zs)
+    isnothing(start_index) && return zs
+    _run(start_index:n, threaded, progress, "Applying $f to each geometry...") do i
         zs[i] = _zonal(f, x, geoms[i]; kw...)
     end
     return zs
 end
 
-function _alloc_zonal(f, x, geom, n; kw...)
-    z1 = _zonal(f, x, geom; kw...)
+function _alloc_zonal(f, x, geoms, n; kw...)
+    # Find first non-missing entry
+    n_missing::Int = 0
+    z1 = _zonal(f, x, first(geoms); kw...)
+    for geom in geoms
+        z1 = _zonal(f, x, geom; kw...)
+        if !ismissing(z1)
+            break
+        end
+        n_missing += 1
+    end
     zs = Vector{Union{Missing,typeof(z1)}}(undef, n)
-    zs[1] = z1
+    zs[1:n_missing] .= missing
+    if n_missing == n
+        return zs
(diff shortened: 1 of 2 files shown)


src/methods/zonal.jl
```diff
@@ -108,16 +108,31 @@ function _zonal(f, x::RasterStackOrArray, ::Nothing, data; progress=true, thread
...
+        end
+        n_missing += 1
+    end
     zs = Vector{Union{Missing,typeof(z1)}}(undef, n)
-    zs[1] = z1
+    zs[1:n_missing] .= missing
+    if n_missing == n
+        return zs
+    end
+    zs[n_missing + 1] = z1
```
rafaqz, line 136: Some comments explaining what this is for would help

src/methods/zonal.jl
```diff
@@ -108,16 +108,31 @@ function _zonal(f, x::RasterStackOrArray, ::Nothing, data; progress=true, thread
     geoms = _get_geometries(data, geometrycolumn)
     n = length(geoms)
     n == 0 && return []
-    zs = _alloc_zonal(f, x, first(geoms), n; kw...)
-    _run(1:n, threaded, progress, "Applying $f to each geometry...") do i
+    zs = _alloc_zonal(f, x, geoms, n; kw...)
+    start_index = findfirst(x -> !ismissing(x), zs)
```
rafaqz, line 112: Can we just return `start_index` from `_alloc_zonal` (e.g. `nmissing + 1`) to avoid the `findfirst` ?

DanTanAtAims: That makes more sense, I've added that to the most recent commit.

src/methods/zonal.jl
DanTanAtAims: I've added clarifications, feel free to indicate more locations I can add clarity.

rafaqz:
This is good, just needs a test

test/methods.jl
```diff
@@ -299,6 +299,16 @@ end
         zonal(sum, st; of=dims(st)) == 
         zonal(sum, st; of=Extents.extent(st)) == 
         sum(st)
+
```
rafaqz, line 302: Thanks! My last nitpick is to put all of this in a `@testset` with a suitable name

DanTanAtAims: I've separated them into a test set called `zonal_return_missing`. Feel free to suggest a better name, naming is hard!

test/methods.jl
```diff
@@ -301,6 +301,18 @@ end
         sum(st)
 end
 
+@testset "zonal_return_missing" begin
```
rafaqz, line 304:
```suggestion
@testset "zonal return missing" begin
```

rafaqz (approved):

PR #716, opened 2024-08-22
define isdisk for AbstractRasterSeries

tiemvanderdeure:
`isdisk` was not defined for AbstractRasterSeries and therefore defaulted to `false`, making `combine` incredibly slow for lazy `AbstractRasterSeries`.

In this PR it just looks at the first raster in the series, which I'm sure there are some edge cases where it doesn't work, but `AbstractRasterStack` does the very same thing.

We might want to define things like `metadata` and `name` as well?

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -24,6 +24,7 @@ abstract type AbstractRasterSeries{T,N,D,A} <: AbstractDimArray{T,N,D,A} end
 DD.metadata(A::AbstractRasterSeries) = NoMetadata()
 DD.name(A::AbstractRasterSeries) = NoName()
 DD.label(A::AbstractRasterSeries) = ""
+isdisk(A::AbstractRasterSeries) = isdisk(first(A))
 
 """
     modify(f, series::AbstractRasterSeries)
--- a/src/series.jl
+++ b/src/series.jl
@@ -21,10 +21,10 @@ abstract type AbstractRasterSeries{T,N,D,A} <: AbstractDimArray{T,N,D,A} end
 
 # Interface methods ####################################################
 
-DD.metadata(A::AbstractRasterSeries) = NoMetadata()
-DD.name(A::AbstractRasterSeries) = NoName()
+DD.metadata(A::AbstractRasterSeries) = DD.metadata(first(A))
+DD.name(A::AbstractRasterSeries) = DD.name(first(A))
 DD.label(A::AbstractRasterSeries) = ""
-isdisk(A::AbstractRasterSeries) = any(isdisk, A))
+isdisk(A::AbstractRasterSeries) = any(isdisk, A)
 
 """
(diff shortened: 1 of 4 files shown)


src/series.jl
```diff
@@ -24,6 +24,7 @@ abstract type AbstractRasterSeries{T,N,D,A} <: AbstractDimArray{T,N,D,A} end
 DD.metadata(A::AbstractRasterSeries) = NoMetadata()
 DD.name(A::AbstractRasterSeries) = NoName()
 DD.label(A::AbstractRasterSeries) = ""
+isdisk(A::AbstractRasterSeries) = isdisk(first(A))
```
rafaqz, line 27:
```suggestion
isdisk(A::AbstractRasterSeries) = any(isdisk, A))
```

tiemvanderdeure:
That was also my first instinct, but then I couldn't figure out if it'd be better to use `any` or `all`. In the unusual case of a series/stack that is party on disk and partly in memory, would it be better to treat it as if it were on disk or in memory?

Or should isdisk return something else (maybe `nothing`) in that case?

src/series.jl
```diff
@@ -21,9 +21,10 @@ abstract type AbstractRasterSeries{T,N,D,A} <: AbstractDimArray{T,N,D,A} end
 
 # Interface methods ####################################################
 
-DD.metadata(A::AbstractRasterSeries) = NoMetadata()
-DD.name(A::AbstractRasterSeries) = NoName()
+DD.metadata(A::AbstractRasterSeries) = DD.metadata(first(A))
+DD.name(A::AbstractRasterSeries) = DD.name(first(A))
```
tiemvanderdeure, line 25: I don't know if this really is ever relevant. But after `combine`, the name/metadata of the first raster are preserved, so this at least is consistent.

src/series.jl
```diff
@@ -21,10 +21,10 @@ abstract type AbstractRasterSeries{T,N,D,A} <: AbstractDimArray{T,N,D,A} end
...
```
rafaqz, line 25: Not sure about these... maybe for a different PR

src/series.jl
rafaqz: Its a bit weird... it makes sense after combine but not before it.

rafaqz:
A test would be nice

test/series.jl
```diff
@@ -122,5 +122,7 @@ end
         @test all(Rasters.filename.(series) .== filenames)
         first_dims = dims(first(series))
         @test all(dims(r) == first_dims for r in series)
-    end
+        @test Rasters.isdisk(series)
```
rafaqz, line 125:
```suggestion
        @test Rasters.isdisk(series)
        @test !Rasters.isdisk(read(series))
```

PR #718, opened 2024-09-01
ditch SurfaceLikeCompat to remove method ambiguity

asinghvi17:
I thought I opened an issue about this but couldn't find it.  This just squashes a method ambiguity which was previously disguised because DDMakieExt and RastersMakieExt both override the common Makie plot functions.  This failed when I was trying to use `GeoMakie.meshimage` with a raster.

### DIFF
--- a/ext/RastersMakieExt/plotrecipes.jl
+++ b/ext/RastersMakieExt/plotrecipes.jl
@@ -294,9 +294,17 @@ end
 function Makie.convert_arguments(t::Makie.PointBased, A::AbstractRaster{<:Number,2})
     return Makie.convert_arguments(t, _prepare_dimarray(A))
 end
-function Makie.convert_arguments(t::SurfaceLikeCompat, A::AbstractRaster{<:Any,2})
-    return Makie.convert_arguments(t, _prepare_dimarray(A))
+@static if isdefined(Makie, :SurfaceLike)
+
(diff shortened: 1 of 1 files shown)


ext/RastersMakieExt/plotrecipes.jl
```diff
@@ -294,9 +294,17 @@ end
 function Makie.convert_arguments(t::Makie.PointBased, A::AbstractRaster{<:Number,2})
     return Makie.convert_arguments(t, _prepare_dimarray(A))
 end
-function Makie.convert_arguments(t::SurfaceLikeCompat, A::AbstractRaster{<:Any,2})
-    return Makie.convert_arguments(t, _prepare_dimarray(A))
+@static if isdefined(Makie, :SurfaceLike)
+
+    function Makie.convert_arguments(t::SurfaceLikeCompat, A::AbstractRaster{var"#s115", 2, D}) where {var"#s115", D<:Tuple}
```
rafaqz, line 299:
```suggestion
    function Makie.convert_arguments(t::SurfaceLike, A::AbstractRaster{var"#s115", 2, D}) where {var"#s115", D<:Tuple}
```

asinghvi17: I guess but this is just simpler

PR #730, opened 2024-09-04
adds docs for resample,  MODIS projection

lazarusA:
Ok. I'm adding this because the usual just happened. I needed to do this, and yes, good luck finding docs or examples 😄 . Well, here, a version. It might need a second pass. 

@asinghvi17 it would be also nice to include also your use cases.


docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
+## resample and Projections with ProjString
```
asinghvi17, line 1:
```suggestion
## `resample` and projections with `ProjString`

File: docs/src/resample_proj.md [hunk 2]
```diff
@@ -0,0 +1,124 @@
...
+
+Geospatial datasets will come in different [projections](https://proj.org/en/9.4/operations/projections/index.html) for whaever reasons. Here, we will focus in some of the most used ones, e.g., `MODIS SINUSOIDAL` and `EPSG`, as well as transformations between them.
```
asinghvi17, line 3:
```suggestion
Geospatial datasets will come in different [projections](https://proj.org/en/9.4/operations/projections/index.html) or coordinate reference systems (CRS) for many reasons. Here, we will focus on `MODIS SINUSOIDAL` and `EPSG`, and transformations between them.
```

docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
...
+
+````@example modis
+fig = plot(ras_m; colorrange=(0,100))
+````
+
+### MODIS SINUSOIDAL PROJECTION
+
+````@example
+# ? is this the right ProjString ?, do we need to shift lat, lon?
+SINUSOIDAL_CRS = ProjString("+proj=sinu +lon_0=0 +x_0=0 +y_0=0 +a=6371007.181 +b=6371007.181 +units=m +no_defs")
```
asinghvi17, line 32: Users will be wondering what the hell a ProjString is, we should explain that

docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
...
+````
+
+now, let's go back to `latitude` and `longitude`
+
+````@example modis
+# ? also here, do we need to shift X, Y?
+ras_epsg = resample(ras_sin; size=(1440,720), crs=EPSG(4326), method="sum")
+````
+
+and let's apply `shiftlocus` such that we can harmonize coordinates, which might be needed when building bigger datasets:
```
asinghvi17, line 60: what does this do, what does it mean to harmonize coordinates?

docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
...
+
+````@example modis
+nansum(ras_m), nansum(locus_resampled)
+````
+
+````@example modis
+fig = plot(ras_epsg; colorrange=(0,100))
+````
+
+### Assemble a Raster from scratch nativily in the sinusoidal projection
```
asinghvi17, line 76:
```suggestion
### Construct a Raster from scratch natively in the sinusoidal projection

File: docs/src/.vitepress/config.mts [hunk 6]
```diff
@@ -73,6 +74,7 @@ export default defineConfig({
       { text: 'Methods',
         items: [
           { text: 'Overview', link: '/methods' },
+          { text: 'resample', link: '/resample_proj'},
```
asinghvi17, line 77: This should be capitalized IMO, or we could call the page `Reprojection` (since that's what it does - resampling to me implies a simple down- or up-sampling without nonlinearity).

asinghvi17 (commented):
I think we need a lot more intro about what exactly this CRS/projection thingy is for new users who might have just loaded their first raster into memory.

docs/src/.vitepress/config.mts
lazarusA: maybe both? `Reprojection and resample`.

docs/src/resample_proj.md
lazarusA: I usually need to match to other variables, which are in a specific Lon/lat range.

asinghvi17:
I'll write a dropdown about what CRS is that we can include here, new users can then click on it if they need to.

rafaqz:
Is this ready to merge?

docs/Project.toml
```diff
@@ -14,7 +14,9 @@ HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
+NaNStatistics = "b946abbf-3ea7-4610-9019-9858bfdeaf2d"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 RasterDataSources = "3cb90ccd-e1b6-4867-9617-4276c8b2ca36"
 Rasters = "a3a2b9e3-a471-40c9-b274-f788e487c689"
 Shapefile = "8e980c4a-a4fe-5da2-b3a7-4b4b0353a2f4"
+YAXArrays = "c21b50f5-aa40-41ea-b809-c0f5e47bfa5c"
```
rafaqz, line 22:
```suggestion
YAXArrays = "c21b50f5-aa40-41ea-b809-c0f5e47bfa5c"
```
```suggestion
```

docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
...
+
+````@example modis
+# ? also here, do we need to shift X, Y?
+ras_epsg = resample(ras_sin; size=(1440,720), crs=EPSG(4326), method="sum")
+````
+
+and let's apply `shiftlocus` such that we can harmonize coordinates, which might be needed when building bigger datasets:
+
+````@example modis
+locus_resampled = DimensionalData.shiftlocus(Center(), ras_epsg)
```
rafaqz, line 63: What happens if we delete these? would be nice to avoid

docs/Project.toml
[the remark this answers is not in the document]
lazarusA: no.

lazarusA:
there some things that might help, like saying or referencing to what `crs` is or a ` ProjString`. Other that, I'm not planning on adding more.

docs/src/.vitepress/config.mts
asinghvi17:
```suggestion
          { text: 'Reprojection and resampling', link: '/resample_proj'},

docs/src/resample_proj.md
asinghvi17: also @lazarusA why are you defining the ellipse here?  That seems like a step too far.  How about:

```julia
SINUSOIDAL_CRS = ProjString("+proj=sinu +lon_0=0 +type=crs")

lazarusA: Those tiles come/are set with very specific numbers. I think I did some tests, and you not always get the same numbers. BTW, feel free to merge this into your PR, I will probably not finish this anytime soon.

lazarusA:
incorporated in https://github.com/rafaqz/Rasters.jl/pull/844

PR #732, opened 2024-09-10
Fix Zarr datasets not picking up CF encoded scale/offset factors

asinghvi17:
This fixes a strange issue I was having where Rasters.jl read the wrong values from a Kerchunk/Zarr dataset.

Based off the `cf` branch, I'm not sure if `get_scale` is added there or is also on the main branch.  Can rebase to main if it doesn't need cf.


rafaqz:
Thanks, probably I've not done this properly for Zarr as it came half way through the `cf` branch work. Iay have missed other things too..

But before and after `cf` are totally different so a PR on `cf` is the way to go.

Thanks for the tests too!

PR #750, opened 2024-09-19
Bump patch version, DD compat

asinghvi17:


rafaqz:
DD compat is a breaking change ;)

PR #762, opened 2024-09-24
Performance improvement by eliminating `map` before  `any` or `all`

tiemvanderdeure:
Improve the performance of a bunch of functions by not calling `map`, but just using `all(f, args)`

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -225,7 +225,7 @@ end
         _dimcoord(d, point)
     end
     # If any are coordinates missing, also return missing for everything
-    if any(ismissing, coords)
+    if any(map(ismissing, coords))
         return missing
     else
         selector_dims = map(dims, coords) do d, c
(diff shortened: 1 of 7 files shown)


rafaqz:
`map` before `all` makes tuples faster but iterables slower, so be careful not to remove it from tuples!

src/methods/extract.jl
```diff
@@ -225,7 +225,7 @@ end
         _dimcoord(d, point)
     end
     # If any are coordinates missing, also return missing for everything
-    if any(map(ismissing, coords))
+    if any(ismissing, coords)
```
rafaqz, line 228: `coords` is a Tuple too

src/methods/extract.jl
```diff
@@ -258,7 +258,7 @@ end
...
```
rafaqz, line 261: And here

tiemvanderdeure:
so it really was just the one in aggregate :)

PR #763, opened 2024-09-25
Breaking: Faster/better `aggregate`

rafaqz:
`aggregate` was written so long ago, was never optimized, and a bunch of things never made sense.

This PR adds optional threading, general performance improvements, and specific fast paths for common methods like `sum` and `mean`. For a `RasterStack` this can be an order of magnitude or more improvement overall.

At the same time I fixed the dumb `dissaggregate` arguments.

@tiemvanderdeure this was inspired by your comment to use `aggregate` proving to be actually slower than `resample`. So if you want to review...


src/methods/aggregate.jl
```diff
@@ -136,40 +142,36 @@ function aggregate!(loci::Tuple{Locus,Vararg}, dst::AbstractRaster, src, scale;
...
     intscale = _scale2int(Ag(), dims(src), scale)
-    broadcast!(dst, CartesianIndices(dst)) do I
+    # len = prod(intscale)
+    l = upsample.(map(firstindex, axes(dst)), intscale)
+    u = upsample.(map(lastindex, axes(dst)), intscale)
+    checkbounds(src, l...)
+    checkbounds(src, u...)
+    # If a disk array, cache the src so we don't read too many times
+    src_parent = isdisk(src) ? DiskArrays.cache(parent(src)) : parent(src)
+    @inbounds broadcast!(dst, CartesianIndices(dst)) do I
```
tiemvanderdeure, line 159: would it be worthwhile to optionally run this threaded?

src/methods/aggregate.jl
```diff
@@ -187,49 +189,55 @@ Disaggregate array, or all arrays in a stack or series, by some scale.
 
 # Keywords
 
-- `progress`: show a progress bar.
+$FILENAME_KEYWORD
+$SUFFIX_KEYWORD
+$PROGRESS_KEYWORD
+$THREADED_KEYWORD
 
 Note: currently it is faster to aggregate over memory-backed arrays.
```
tiemvanderdeure, line 197: should this say disaggregate?

tiemvanderdeure (approved):

src/methods/aggregate.jl
```diff
@@ -5,42 +5,53 @@ const IntOrIntTuple = Union{Int,Tuple{Vararg{<:Int}}}
 struct Ag end
 struct DisAg end
 
+const SKIPMISSING_KEYWORD = """
+- `skipmissing`: if `true`, any `missingval` will be skipped during aggregation, so that
+    only areas of all missing values will be aggregated to `missingval`. If `false`, any
+    aggregated area containing a `missingval` will be assigned `missingval`.
+"""
```
tiemvanderdeure, line 12: add "false by default"

src/methods/aggregate.jl
rafaqz: Oh the reason not to is currently it works on GPU and DiskArrays as is, threading will break that

src/methods/aggregate.jl
rafaqz: It can be deleted, the caching should make it fast now

src/methods/aggregate.jl
```diff
@@ -259,10 +264,10 @@ Disaggregate array `src` to array `dst` by some scale, using `method`.
...
 Use [`read`](@ref) on `src` before use where required.
 """
-function disaggregate!(locus::Locus, dst::AbstractRaster, src, scale)
+function disaggregate!(dst::AbstractRaster, src, scale)
     disaggregate!((locus,), dst, src, scale)
 end
-function disaggregate!(loci::Tuple{Locus,Vararg}, dst::AbstractRaster, src, scale)
+function disaggregate!(dst::AbstractRaster, src, scale)
     intscale = _scale2int(DisAg(), dims(src), scale)
     broadcast!(dst, CartesianIndices(dst)) do I
```
tiemvanderdeure, line 272: It must be faster to loop through the `src` instead like so since we know it is smaller.

By looping through `dst` we have lots of unnecessary calls to `upsample` and lookups in `src`.

```suggestion
    for I in CartesianIndices(src)
        upper = upsample.(Tuple(I), intscale)
        lower = upper .+ intscale .- 1
        I_dst = map(:, upper, lower)
        val = src[I]
        val === missingval(src) ? missingval(dst) : val
        view(dst, I_dst...) .= val
    end
    return dst
```

tiemvanderdeure:
Amazing! Did a few quick tests/profviews and it seems blazing fast - I can't come up with a way to improve there. I just had a quick thing for disaggregate.

A few other suggestions (short or long term) - now that we're making considerable changes to this function anyway: 
- would it make sense to implement aggregate for `RasterSeries`? 
- would it make sense to only aggregate along contiuous dimensions by default. So that a raster like `Raster(rand(X(1:10), Band(["a", "b"])` won't get aggregated along the Band dimension unless requested explicitly?
- Would we want an `interpolate` function that works like `disaggregate` but does more than just copy values over. I guess this was the idea with the function argument in disaggregate in the first place?

src/methods/aggregate.jl
```diff
@@ -259,10 +264,10 @@ Disaggregate array `src` to array `dst` by some scale, using `method`.
 Note: currently it is faster to aggregate over memory-backed arrays.
 Use [`read`](@ref) on `src` before use where required.
 """
-function disaggregate!(locus::Locus, dst::AbstractRaster, src, scale)
+function disaggregate!(dst::AbstractRaster, src, scale)
```
tiemvanderdeure, line 267: Don't understand this one - Isn't this the exact same type signature as the method just below?

src/methods/aggregate.jl
tiemvanderdeure: Can't comment on unchanged lines, but if we implement this then `downsample` isn't used anywhere anymore.

rafaqz:
>    would it make sense to implement aggregate for RasterSeries?
Its there! I also threaded it

>    would it make sense to only aggregate along contiuous dimensions by default. So that a raster like Raster(rand(X(1:10), Band(["a", "b"]) won't get aggregated along the Band dimension unless requested explicitly?

You mean `AbstractSampled` lookups are aggregated by `Cetegorical` are not? Yes I thought about that too looking at the R code doing that in the benchmarks. We basically never want to aggegate adjacent categories. You would use `groupby` for something like that.

>    Would we want an interpolate function that works like disaggregate but does more than just copy values over. I guess this was the idea with the function argument in disaggregate in the first place?

Yeah it was going to be but never happened lol. Probably it should be another geostatistics gap fillin method instead of here.

src/methods/aggregate.jl
rafaqz: I forgot to delete a method

src/methods/aggregate.jl
rafaqz: Right yes it will be faster. I think the hidden reason again is that wont work on a GPU or a disk array and I was trying to be generic

rafaqz: We could use Flatten.jl so see if there is an array inside whatever wrappers, and if it is do the loop, if not the broadcast

rafaqz: We need DiskArraysKernelAbstractions then we could just do all of this with KernelAbstractions, and use Stencils.jl too to make it even faster

tiemvanderdeure:
> You mean AbstractSampled and NoLookup lookups are aggregated but Categorical are not?

Yes I think that could be the default if `scale` is an integer. And to override it users can always provide a tuple and specify.

It could also depend on the order - so we don't aggregate along `Unordered` dimensions.

src/methods/aggregate.jl
tiemvanderdeure: What about this woulnd't work with on GPU? We're just viewing into an array and filling it with some value, right?

rafaqz: Its all scalar indexing. We need to launch single kernel to do the whole lot in one go.

rafaqz: broacasts handle kernel launches for us. If you want to write manual code like that with blocks with need KernelAbstractionsjl

tiemvanderdeure: Another idea is to do some clever reshapes and permutes to be able to broadcast without having to do indexing math ourselves. I tried this implementation and it works, but it's actually a little slower than the original. I guess the reshaping makes indexing that much slower.

 ```julia
   intscale = _scale2int(DisAg(), dims(src), scale)
    n = length(intscale)
    reshape_dims = ntuple(i -> mod(i, 2) == 0 ? size(src)[(i ÷ 2)] : intscale[(i ÷ 2) + 1], Val(n*2))
    permute_order = (range(2; step = 2, length = n)..., range(1, step = 2, length = n)...)
    dst_reshaped = Base.PermutedDimsArray(Base.reshape(dst, reshape_dims), permute_order)
    broadcast!(dst_reshaped, parent(src)) do val
        val === missingval(src) ? missingval(dst) : val
    end
    return dst
```

src/methods/aggregate.jl
tiemvanderdeure: well if threaded is false by default then you could either broadcast or do a threaded for loop. But after seeing just how fast this is I don't think that would only be relevant in some niche use cases

src/methods/aggregate.jl
rafaqz: Actually no `cache` is broken on our DiskArrays version 😭

rafaqz:
Ok I've implemented this so you have to explicitly aggregate categorical dimensions. I'm not sure what to do with the categories? Maybe we need to join as strings somehow? Or return `NoLookup` ?

With `Unordered` lookups we really can't aggregate them in a sensible way - I guess if people explicitly force it we can just return `NoLookup` too, to show that the lookups are effectively destroyed by the process?

rafaqz:
@tiemvanderdeure want to do one last review of this? would be good to get it in with the other breaking changes

tiemvanderdeure:
I think there's a commit missing here. The behaviour with unordered or missing lookups is not implemented or tested

tiemvanderdeure:
But you said this might be on your old laptop, right?

rafaqz:
Oh no yes it probably is. I'll have to put the NVME in some other computer

tiemvanderdeure:
Bump! Did you get your data back from the broken laptop?

rafaqz:
Ugh not yet, but I have it on an external drive at least. Will do it soon

Issue #764, opened 2024-09-25
cellsize returns erroneous results for grids with fine spacing [no reprojeciton]


alex-s-gardner:
This makes sense as gridsize is increasing linearly with latitude
```
using Rasters
using Rasters.Lookups
import GeoFormatTypes as GFT
using Plots

dX = 0.1
dY = -0.1
lon = X(Projected(166.:dX:168.; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dX), crs=EPSG(4326)))
lat = Y(Projected(-78.0:dY:-80.; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dY ), crs=EPSG(4326)))
ras = Raster(rand(lon, lat))
cell_area = Rasters.cellsize(ras)
plot(cell_area[1, :])
```
<img width="431" alt="Screenshot 2024-09-25 at 1 19 39 PM" src="https://github.com/user-attachments/assets/2bee944e-35d1-4a85-bd91-8221bf514ff1">

if we decrease the step size in X we start to see some jutter
```
dX = 0.0006
dY = -0.1
lon = X(Projected(166.0:dX:168.0; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dX), crs=EPSG(4326)))
lat = Y(Projected(-78.0:dY:-80.0; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dY), crs=EPSG(4326)))
ras = Raster(rand(lon, lat))
cell_area = Rasters.cellsize(ras)
plot(cell_area[1, :])
```
<img width="429" alt="Screenshot 2024-09-25 at 1 24 47 PM" src="https://github.com/user-attachments/assets/10ac7056-3841-4c94-b2e5-004a310360ba">

and things really fall apart when we decrease step size in X and Y
```
dX = 0.0006
dY = -0.0003
lon = X(Projected(166.0:dX:167.0; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dX), crs=EPSG(4326)))
lat = Y(Projected(-78.0:dY:-79.0; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dY), crs=EPSG(4326)))
ras = Raster(rand(lon, lat))
cell_area = Rasters.cellsize(ras)
plot(cell_area[1, :])
```
[...]

Issue #747, opened 2024-09-19
recommendation: change `cellsize` to `cellarea`


alex-s-gardner:
cellsize is a bit ambiguous, cell area is not

Also, it could be helpful if cellsize() accepted a kwarg of "unit" if the user wanted a unit other than km^2 returned. 

Given that projected units are always in meters, and meter is si units, I wonder if returning meters^2 by default would be a better choice in the long run.

rafaqz:
I'll add unitful support soon for CF as well, so that can happen at the same time.

For `cellarea` PR and deprecated `cellsize`?

@tiemvanderdeure ?

tiemvanderdeure:
I have no strong feelings about this one. The name cellsize comes from the function in terra in R with the same name: https://rdrr.io/cran/terra/man/cellSize.html. 

Changing to meter units would be breaking but it would make sense to do it at the same time as a name change.

rafaqz:
I think we will have a `units` keyword for opt-in units wherever they're used, `false` by default.

PR #768, opened 2024-09-27
improve cellsize and switch from Archgdal to Proj

tiemvanderdeure:
Solves https://github.com/rafaqz/Rasters.jl/issues/764 and makes cellsize calculation much faster for lon-lat projections. 

I'm considering getting rid of the ArchGDAL requirement for these projections as well. 

Any ideas for further improvements @rafaqz @alex-s-gardner?


rafaqz:
How do we get rid of ArchGDAL?

tiemvanderdeure:
I was thinking we could just make a `_crs2transform` that just calls `ArchGDAL.crs2transform` if ArchGDAL is loaded and otherwise tells the user to load ArchGDAL. If the CRS is WGS84 then `cellsize` would never have to call `_crs2transform`

rafaqz:
But how do we deal with Proj/Wkt, a lot of different text strings can be wgs84

tiemvanderdeure:
> But how do we deal with Proj/Wkt, a lot of different text strings can be wgs84

The way we already do it - I don't think this line of code needs ArchGDAL

https://github.com/rafaqz/Rasters.jl/blob/8d13c1887f90444d24e3c19ecbff42c8d5f0de68/ext/RastersArchGDALExt/cellsize.jl#L57

rafaqz:
Yeah that's ArchGDAL doing that conversion ;)

ext/RastersArchGDALExt/cellsize.jl
```diff
@@ -47,25 +47,30 @@ function _area_from_coords(transform::ArchGDAL.CoordTransform, ::GI.LinearRingTr
...
+    two_pi_R2 = 2 * pi * R * R
+    band_area = broadcast(DD.intervalbounds(lat)) do yb
+        two_pi_R2 * (sin(deg2rad(yb[2])) - sin(deg2rad(yb[1])))
+    end
+    
+    xwidths = broadcast(DD.intervalbounds(lon)) do xb
+        abs(xb[2] - xb[1]) / 360
+    end
+    
+    xwidths .* band_area'
```
rafaqz, line 61: We can combine these broadcasts to remove one allocation

tiemvanderdeure:
> Yeah that's ArchGDAL doing that conversion ;)
I was thrown off by the GeoFormatTypes wrapper, but makes sense. Let's just leave it as it is, then.

rafaqz:
Yes ArchGDAL actually pirates `Base.convert` on `GeoFormat`, so the consfusion is understandable. We could put it in a GeoFormatTypesArchGDAL extension now, but that wasn't possible at the time. We could also switch to using Proj.jl directly.

alex-s-gardner:
Can we take this opportunity to rename `cellsize` to `cellarea` and to return area in units of meters #747, also adding a `units` kwarg?

alex-s-gardner:
There is room for significant speed gains for small small grid spacing:

```
#For lat/lon aligned rasters with small grid spacing, this will be much faster and just as accurate:

# load packages
import Rasters: EPSG
using Rasters
using Rasters.Lookups
using DimensionalData


# define meters to lat/lon function
"""
    meters2lonlat_distance(distance_meters, latitude_degrees)

Returns the decimal degree distance along latitude and longitude lines given a distance in 
meters and a latitude in decimal degrees.

# Example usage:
```julia-repl
julia> distance_meters = 1000.0;  
julia> latitude_degrees = 45.0;  

julia> lat, lon = Altim.meters2lonlat_distance(distance_meters, latitude_degrees)
(0.008997741566866717, 0.012718328120254203)
```
"""
function meters2lonlat_distance(distance_meters, latitude_degrees)
    # Radius of the Earth in meters
    earth_radius = 6371000

    # Calculate the angular distance in radians
    angular_distance = distance_meters / earth_radius

    # Calculate the longitude distance using the Haversine formula
    longitude_distance = angular_distance * (180.0 / π) / cosd(latitude_degrees)

    latitude_distance = distance_meters / 111139

    return latitude_distance, longitude_distance
end

# build an example raster
dX = 0.1
dY = -0.1
lon = X(Projected(166.:dX:168.; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dX), crs=EPSG(4326)))
lat = Y(Projected(-78.0:dY:-80.; sampling=Intervals(Start()), order=ForwardOrdered(), span=Regular(dY ), crs=EPSG(4326)))
ras = Raster(rand(lon, lat))

# given a lat/lon raster with small grid spacing caclculate area
dlon = dims(ras, :X)
dlat = dims(ras, :Y)
lonlat_per_meter = meters2lonlat_distance.(Ref(1), dlat)

dist_lon = step(dlon) ./ getindex.(lonlat_per_meter, 1)
dist_lat = step(dlat) ./ getindex.(lonlat_per_meter, 2)
area = ones(dlon) * DimArray(dist_lon .* dist_lat, dlat)'
```

**Comment by rafaqz** [comment]
> Can we take this opportunity to rename cellsize to cellarea and to return area in units of meters #747, also adding a units kwarg?

Let's wait for a complete Unitful extension to add the `units` keyword to everything all at once. (in a few months when my PhD is done 😅)

I'm happy to rename but we should `@deprecate` the current name so it still works for a while.

**Comment by alex-s-gardner** [comment]
> I'm happy to rename but we should `@deprecate` the current name so it still works for a while.

given the incorrectness of `cellsize` it might actually be better to break it

**Comment by tiemvanderdeure** [comment]
> There is room for significant speed gains for small small grid spacing:

Have you compared to the implementation in this PR? For me it is giving similar speeds.

I can see the point of your implementation, but using the haversine formula also has downsides, since it doesn't account for the curvature of the earth. I know this isn't a big deal in most cases, but the error increases with the size of gridcells. So after (dis)aggregating a raster, the total area returned by cellarea would be different, which is not ideal. 

Maybe we should implement it for other projections than lon-lat, but then we might need to do some more geometry.

**Comment by tiemvanderdeure** [comment]
> > Can we take this opportunity to rename cellsize to cellarea and to return area in units of meters #747, also adding a units kwarg?
> 
> Let's wait for a complete Unitful extension to add the `units` keyword to everything all at once. (in a few months when my PhD is done 😅)

Maybe we should already start returning the result in metres though, to avoid another breaking change down the line. If we add a units keyword later then it would default to m^2.

**Comment by rafaqz** [comment]
Yeah result in metres is fine, just not the units kw

**Comment by alex-s-gardner** [comment]
> Have you compared to the implementation in this PR? For me it is giving similar speeds.

Your implementation knocks the socks off of my proposal (2x faster) and is more accurate so ignore my recommendation. Excited to see this progressing... fast, useful and intuitive... couldn't ask for anything more

**Comment by rafaqz** [comment]
Is this good to go?

**Comment by tiemvanderdeure** [comment]
I think so! Can you or @asinghvi17  just nod at the logic in the line of code that reprojects to `mappedcrs`? I don't think we need an option to turn that behaviour off, right? I guess users vouch for the accuracy of that transformation by providing a `mappedcrs`.

File: ext/RastersArchGDALExt/cellarea.jl [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+    end
+    
+    return area*radius^2
+end
+
+_area_from_coords(transform, geom; radius) = _area_from_coords(transform, GI.trait(geom), geom; radius)
+function _area_from_coords(transform::ArchGDAL.CoordTransform, ::GI.LinearRingTrait, ring; radius)
+    points = map(GI.getpoint(ring)) do p 
+        t = ArchGDAL.transform!(ArchGDAL.createpoint(p...), transform)
+        (deg2rad(GI.x(t)), deg2rad(GI.y(t)))
```
asinghvi17, line 38: If we use `cosd` and `sind` the need for this goes away.  Basic tests indicate that it's not too inaccurate but I'd rather be safe.

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
+function _great_circle_bearing(lon1::AbstractFloat, lat1::AbstractFloat, lon2::AbstractFloat, lat2::AbstractFloat)
+    dLong = lon1 - lon2
+
+    s = cos(lat2)*sin(dLong)
+    c = cos(lat1)*sin(lat2) - sin(lat1)*cos(lat2)*cos(dLong)
```
asinghvi17, line 5: These could all be cosd or sind, so we might not need deg2rad in the invoking method.

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+        (deg2rad(GI.x(t)), deg2rad(GI.y(t)))
+    end
+    return _area_from_rads(GI.LinearRing(points); radius)
+end
+
+# For lat-lon projections. Get the area of each latitudinal band, then multiply by the width
+function _area_from_lonlat(lon::XDim, lat::YDim; radius)
+    two_pi_R2 = 2 * pi * radius * radius
+    band_area = broadcast(DD.intervalbounds(lat)) do yb
+        two_pi_R2 * (sin(deg2rad(yb[2])) - sin(deg2rad(yb[1])))
```
asinghvi17, line 47: We could also use `cosd` and `sind` here, they're a hair more accurate

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+_isgeographic(crs) = _isgeographic(ArchGDAL.importCRS(crs))
+_isgeographic(crs::AG.ISpatialRef) = AG.GDAL.osrisgeographic(crs) |> Bool
+
+_isdegrees(crs) = _isdegrees(ArchGDAL.importCRS(crs))
+function _isdegrees(crs::AG.ISpatialRef)
+    _isgeographic(crs) || return false
+    pointer = Ref{Cstring}()
+    result = AG.GDAL.osrgetangularunits(crs, pointer)
+    return unsafe_string(pointer[]) == "degree"
+end
```
asinghvi17, line 98: Nice!

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+function cellarea(x::Union{<:AbstractRaster, <:AbstractRasterStack, <:RA.DimTuple}; kw...)
+    cellarea(dims(x, (XDim, YDim)); kw...)
+end
+
+# TODO: put these in ArchGDAL
+_isgeographic(crs) = _isgeographic(ArchGDAL.importCRS(crs))
+_isgeographic(crs::AG.ISpatialRef) = AG.GDAL.osrisgeographic(crs) |> Bool
+
+_isdegrees(crs) = _isdegrees(ArchGDAL.importCRS(crs))
+function _isdegrees(crs::AG.ISpatialRef)
```
asinghvi17, line 93: Why the mix between `ArchGDAL` and `AG`?  Could we stick to one?  It looks like `ArchGDAL` is used everywhere else and it's not too long either ;)

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+    
+    broadcast(DD.intervalbounds(lon), band_area') do xb, ba
+        abs(xb[2] - xb[1]) / 360 * ba
+    end
+end
+
+function cellarea(dims::Tuple{<:XDim, <:YDim}; radius = 6371008.8)
+    # check the dimensions 
+    isnothing(crs(dims)) && _no_crs_error()
+    any(d -> d isa Points, sampling.(dims)) && throw(ArgumentError("Cannot calculate cell size for a `Raster` with `Points` sampling."))
```
asinghvi17, line 58: Should this be either `ispoints` or `!isintervals` instead?

ext/RastersArchGDALExt/cellarea.jl
asinghvi17:
```

julia> _isdegrees(GFT.EPSG(3857))
false

julia> _isgeographic(GFT.EPSG(3857))
false

```

The code here is web mercator, which is known remappable to long/lat.  So we might need another check...

asinghvi17:
Feel free to ignore the cosd/sind stuff if there's no time, I just discovered that they use an extended-precision deg2rad that might be useful.

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    areas = if _isdegrees(crs(dims)) # check if need to reproject
+        _area_from_lonlat(dims...; radius)
+    elseif !isnothing(mappedcrs(dims)) && _isdegrees(mappedcrs(dims))
```
asinghvi17, line 62: Any mapped crs will be in meters though?  So isdegrees seems kind of irrelevant

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,98 @@
...
+        _area_from_lonlat(reproject(dims; crs = mappedcrs(dims))...; radius)
```
asinghvi17, line 63: Shouldn't the crs be `EPSG(4326)` in this line?

test/cellarea.jl
```diff
@@ -0,0 +1,59 @@
...
+    lat = Lat(Projected(0.0:0.1:89.9; sampling=Intervals(Start()), order = ForwardOrdered(), span = Regular(0.1), crs=EPSG(4326)))
+    cs_lat = cellarea((x, lat))
+    @test parent(cs_lat) == parent(cs)
+    @test dims(cs_lat) == (x, lat)
+
+    # test using mapped dims to calculate cellsize
+    x_3857 = X(Projected(0:1e5:1e7; sampling=Intervals(Start()), order = ForwardOrdered(), span = Regular(1e5), crs=EPSG(3857), dim = X()))
+    y_3857 = Y(Projected(0:1e5:1e7; sampling=Intervals(Start()), order = ForwardOrdered(), span = Regular(1e5), crs=EPSG(3857), dim = Y()))
+    dimz_3857 = (x_3857, y_3857)
+    mappeddimz = setmappedcrs(dimz_3857, EPSG(4326))
```
asinghvi17, line 48: I think it's most commonly the other way around?  

Where you have a CRS that _can_ be remapped to lat/long, we should check if it can be done and then compute cell area through the fast path.

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: But it's still in meters so we don't want to use the _area_from_lonlat for that projection. I tried for EPSG(4326) and EPSG(4258) which both returned true. It's the best Raf and I could come up with and I think it works, but the documentation on these functions is pretty sparse and there are just so many cases.

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: I didn't know about those, thanks!

ext/RastersArchGDALExt/cellarea.jl
asinghvi17: True, that requires users to set the mapped crs before invoking cellarea though.  

I guess this can go in another PR.  My thought was that it would look something like this:
- start out with a lat/long raster
- set mappedcrs to local UTM so that life is easier
- try to compute cellarea, it takes a while

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: the idea is that we just checked that mappedcrs is a projection that uses degrees, so we can use it whatever it is

ext/RastersArchGDALExt/cellarea.jl
asinghvi17: https://github.com/rafaqz/Rasters.jl/blob/8d13c1887f90444d24e3c19ecbff42c8d5f0de68/ext/RastersArchGDALExt/reproject.jl#L7-L9

is the approximate check that's currently being used I think, we could make it a bit more robust

ext/RastersArchGDALExt/cellarea.jl
rafaqz: I think it was originally written before they existed, or at a similar time

ext/RastersArchGDALExt/cellarea.jl
rafaqz: `mappedcrs` will most often be in degrees

ext/RastersArchGDALExt/cellarea.jl
rafaqz: Yeah it might be other projections in degrees, they will still work

src/extensions.jl
```diff
@@ -175,25 +176,35 @@ using Rasters, ArchGDAL, Rasters.Lookups
...
+  90.0  1.23017e6   1.19279e6   1.11917e6   1.01154e6  873182.0  708290.0
+ 100.0  1.23017e6   1.19279e6   1.11917e6   1.01154e6  873182.0  708290.0
+ 110.0  1.23017e6   1.19279e6   1.11917e6   1.01154e6  873182.0  708290.0
+ 120.0  1.23017e6   1.19279e6   1.11917e6   1.01154e6  873182.0  708290.0
 ```
 $EXPERIMENTAL
 """
-cellsize(args...; kw...) = throw_extension_error(cellsize, "ArchGDAL", :RastersArchGDALExt, args)
+cellarea(args...; kw...) = throw_extension_error(cellarea, "ArchGDAL", :RastersArchGDALExt, args)
+function cellsize(args...; kw...)
```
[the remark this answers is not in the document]
tiemvanderdeure: I did this manually to 1) make sure this is displayed (by default it is not), and 2) inform that the unit is changed.

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: If the raster is in lat/long then cellarea will be fast no matter what the mappedcrs is, because we check the crs first. And if the lat/lon raster got warped, then using the original lat/lon system won't be accurate. We can't in all cases project the dims to EPSG(4326), and if we don't know that we can, we should project the points instead.

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: thanks, I  changed it to `isintervals(dims) || throw`

test/cellarea.jl
tiemvanderdeure: With "the other way around" do you mean cases where the crs is lat/lon and the mappedcrs is something different? In that case cellarea will use the crs.

tiemvanderdeure: cellarea will only use mappedcrs if the crs is not in degrees, but mappedcrs is, so that's why the test is this way around.

ext/RastersArchGDALExt/cellarea.jl
asinghvi17: Gotcha, that makes sense.

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: Maybe we should keep this for a separate PR, since it's a bigger discussion. If we've missed any cases where we're not using the shortcut for lon/lat projections, then cellarea will use the fallback which is still really accurate (except for very small gridcells)

tiemvanderdeure:
Thanks for reviewing @asinghvi17! I didn't know about cosd and sind so that was really helpful.

I think this is really solid now! The only thing I could think of that we could still add is an option to disable the transformation and just return the cartesian area for planar projections. Should I add it?

ext/RastersArchGDALExt/cellarea.jl
alex-s-gardner: If the fallback is what we previously had then I would say we need a better solution. It's pretty common for geospatial datasets to have pixel spacings of < 10 m.. but maybe I misunderstood.

alex-s-gardner:
> I think this is really solid now! The only thing I could think of that we could still add is an option to disable the transformation and just return the cartesian area for planar projections. Should I add it?

I think this would be super useful. If the user understands their projection and just want map ptojected area in map units they could simply pass a kwarg like `area_in_crs` or similar

ext/RastersArchGDALExt/cellarea.jl
tiemvanderdeure: The fallback is still to use Girard's theorem, which is what we had previously. It might be slightly more accurate now that we use sind and cosd, but for such small pixel spacings my guess is that it will still suffer from numerical stability issues. But I don't really know what to use instead

rafaqz: > If the fallback is what we previously had then I would say we need a better solution. 

What is you idea for this @alex-s-gardner ?

alex-s-gardner: The previous precision errors were pretty strange if everything was being maintained at Float64. Maybe the issue was the sind and cosd and then it's a non-issue. If not then I guess we fallback to a simple width * height for very small gidcells? Though that seems like a hack.

tiemvanderdeure: Using Girard's theorem means we sum the angles of the gridcell and see how much bigger that is than it would be on a plane. For a 100m gridcell that is in the difference in the angles is in the order of magnitude e-12, and we're doing lots of sin and cos to get there, so I do think it could be a numerical accuracy issue. 

I don't think width * height will be so easy, though. After reprojecting the gridcells won't be rectangular anymore, so we need to calculate a straight angle. And on this scale I don't know if numerical accuracy will then again become an issue?

tiemvanderdeure:
> The only thing I could think of that we could still add is an option to disable the transformation and just return the cartesian area for planar projections.

I implemented this now and it works (even without ArchGDAL). I don't know if I like `area_in_crs` so much, though. terra uses `transform`, but that's also not super clear, is it?

asinghvi17:
Ideally we would say `cellarea(Linear(), raster)` or `cellarea(Spherical(), raster)` I guess, that will have to wait for a bunch of reactors in GeometryOps so that Rasters can depend on a "GeometryOpsCore" that exports these types

alex-s-gardner:
> Ideally we would say `cellarea(Linear(), raster)` or `cellarea(Spherical(), raster)`

It seems like now would be the time to do this as this is a breaking change. We could make it easy and just remove kwarg in favor of Linear() and Spherical() types that could be internally defined... then we can expand in future to be more flexible and to eventually integrate GeometryOps... but at a later date?

asinghvi17:
It would take about 3 days to get out, but we could technically define a GeometryOpsCore.jl package today and get it registered, Rasters could then depend on that (it should have way less dependencies / load time)

ext/RastersArchGDALExt/cellarea.jl
rafaqz: Is there a crossover point where the angles are close enough to 90 degrees that we can switch to just doing h * w?

tiemvanderdeure: Not in any predictable way, right? The size of the gridcell doesn't change the angles.

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,105 @@
...
+    x = cosd(lat) * cosd(lon)
+    y = cosd(lat) * sind(lon)
+    z = sind(lat)
+    return SphericalPoint(x,y,z)
+end
+
+_area_from_coords(transform, geom) = _area_from_coords(transform, GI.trait(geom), geom)
+function _area_from_coords(transform::AG.CoordTransform, ::GI.LinearRingTrait, ring)
+    points = map(GI.getpoint(ring)) do p 
+        t = AG.transform!(AG.createpoint(p...), transform)
```
rafaqz, line 50: `createpoint` is super slow, its allocating C objects that need a finalizer and GC for every single point. Can we run the transformation on larger geometries an skip this? Like just `GI.convert(LinearRing, ring)` instead?

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,105 @@
...
+        (GI.x(t), GI.y(t))
+    end
+    return _spherical_quadrilateral_area(GI.LinearRing(points))
```
rafaqz, line 53:
```suggestion
function _area_from_coords(transform::AG.CoordTransform, ::GI.LinearRingTrait, ring)
    t = AG.transform!(GI.convert(AG.geointerface_geomtype(LinearRingTrait), ring), transform)
    return _spherical_quadrilateral_area(GI.convert(LinearRing, t))
```
Something like this should be better as GI skips around ArchGDAL point creation and just works on the internal vectors of points in the LinearRing

rafaqz: Guessing this will be 3x faster

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,105 @@
...
+        R2 = radius * radius
+        AG.crs2transform(crs(dims), EPSG(4326); order = :trad) do transform
+            [_area_from_coords(
+                transform,         
+                GI.LinearRing([
+                    (xb[1], yb[1]), 
+                    (xb[2], yb[1]), 
+                    (xb[2], yb[2]), 
+                    (xb[1], yb[2]),
+                    (xb[1], yb[1])
```
rafaqz, line 87: This might give some more micro optimisation, 4 things is usually better than 5 things with computers...
```suggestion
                GI.LineSting([
                    (xb[1], yb[1]), 
                    (xb[2], yb[1]), 
                    (xb[2], yb[2]), 
                    (xb[1], yb[2]),
```

ext/RastersArchGDALExt/cellarea.jl
```diff
@@ -0,0 +1,105 @@
...
+dot(p::SphericalPoint, q::SphericalPoint) = sum(p * q)
+function cross(a::SphericalPoint, b::SphericalPoint)
+	a1, a2, a3 = a.data
+    b1, b2, b3 = b.data
+	SphericalPoint((a2*b3-a3*b2, a3*b1-a1*b3, a1*b2-a2*b1))
+end
+
+function _spherical_quadrilateral_area(ring)
+    ps = GI.getpoint(ring)
+    (p1, p2, p3, p4) = _lonlat_to_sphericalpoint.((ps[1], ps[2], ps[3], ps[4]))
```
rafaqz, line 25: The last point isn't actually used here so my `LineString` approach below does make sense

asinghvi17:
Happy to report that this solves #764 
hello

ext/RastersArchGDALExt/cellarea.jl
asinghvi17:
```suggestion                
                GI.LineString([
                    (xb[1], yb[1]), 
                    (xb[2], yb[1]), 
                    (xb[2], yb[2]), 
                    (xb[1], yb[2]),
```

typo

asinghvi17:
Full changelog:
- Switch `cellarea` and `reproject` to use Proj instead of ArchGDAL, increasing speed and decreasing the number of allocations.
- Cellarea changes:
	- Rename `cellsize` to `cellarea`
	- Return output in meters by default (can be overridden by the radius in the manifold, so `Spherical(radius_of_earth_in_km)` gives you are in km, for example.  This could be made more user friendly in the future.
	- Allow "algorithm choice" using GeometryOpsCore Manifold types (planar / spherical)
	- Use [Eriksson's formula for spherical triangle area](https://www.jstor.org/stable/2691141) to compute the area of a grid cell using Cartesian (x, y, z) coordinates for spherical cellarea, increasing precision and providing correct cell area on at least 5mx5m grid cells (could be smaller, haven't tested)
	- Switch to using `sind` and `cosd` rather than `deg2rad`, since they use extended-precision `deg2rad` internally
	- Use a direct method to get cell area for lat-long (geographic) projections, greatly simplifying the process
- Reproject changes:
	- Switch from transforming vectors of points out-of-place with ArchGDAL to transforming a single array of x or y values in-place with Proj's `proj_trans_generic`.  This drastically reduces allocations and memory usage, and improves runtime by 6x.
	- Use Proj's geographic / degree-unit checks to figure out if the CRS is long-lat.  This currently checks explicitly for the "degree" unit name, but this could be expanded in the future.

tiemvanderdeure:
The test failures are unrelated and already fixed in https://github.com/rafaqz/Rasters.jl/pull/780. There is still room for improvement performance-wise but I think we should merge this now. We can make a different PR for further improvements.

src/methods/reproject.jl
```diff
@@ -74,7 +74,7 @@ function _reproject(source::Nothing, target::GeoFormat, dim::Union{XDim,YDim}, v
     reshape(reproject(source, target, dim, vec(vals)), size(vals))
 end
 function _reproject(source::GeoFormat, target::GeoFormat, dim::Union{XDim,YDim}, vals::AbstractVector) 
-    error("Rasters.jl requires backends to be loaded externally as of version 0.8. Run `using ArchGDAL` to use `reproject`")
+    throw_extension_error(_reproject, "Proj", :RastersProjExt, (source, target, dim, vals))
```
rafaqz, line 77:
```suggestion
    throw_extension_error(reproject, "Proj", :RastersProjExt, (source, target, dim, vals))
```

rafaqz (approved):

rafaqz:
Thanks everyone!

alex-s-gardner:
Woot woot ! Fantastic effort.

PR #775, opened 2024-10-02
[WIP] `mapzonal`, allowing you to apply some operation per raster cell, then `zonal` it

asinghvi17:
This PR introduces a function `mapzonal` that looks like this:

```julia
mapzonal(reducer, operator, raster; of = geometries)
```

`operator` is fed each slice of the raster in all dimensions except X and Y.  For a RasterStack that has only `X` and `Y` dimensions, for example, `operator` would be passed the output of `rasterstack[X(ind), Y(ind)]` for each valid index of X and Y.

If you wanted to calculate NDVI from some super-high-fidelity rasterstack, for example, you would perform the computation on demand in `operator`.

Then, `reducer` is passed the vector generated by `operator.(each_cell_of_rasterstack)`.  This can be e.g. `sum`, or even a histogram function.

## Motivation

The motivation here is that `zonal` iterates over the layers of a rasterstack before passing them to the reducer - this can be good, and necessary for the most common workflows, but we needed to operate on the raw values of the individual layers simultaneously.

Some interesting usecases might be:

1. Imagine I have a Raster of soil type and canopy height and I want to know the average canopy height for each soil type within each watershed
2. I have a map of slope-aspect and a map of temperature and I want to know the maximum temperature for each aspect for different mountain ranges
3. I want to calculate the average surface lapse rate (temperature vs elevation) over glacier surfaces given a DEM and raster of surface temperature

## TODOs

- [ ] Tests
- [ ] Comprehensive documentation
- [ ] Performance on regular Rasters with >2 dims (it's currently 10x slower, since indexing into a raster with a dimindex of lacking rank causes a new raster to be allocated)
- [ ] Benchmarking

### DIFF
--- a/src/methods/mapzonal.jl
+++ b/src/methods/mapzonal.jl
@@ -1,0 +1,132 @@
+import Rasters: DD, Extents, GI
+import Rasters: OBJ_ARGUMENT, GEOMETRYCOLUMN_KEYWORD, _get_geometries, _run, RasterStackOrArray
+"""
+    mapzonal(reducer, operator, data::RasterStack; of, kw...)
+
+Think of this like `zonal`, but a named tuple gets sent to the `operator`, 
+and `reducer` gets the result of `operator.(data[mask])`.
+
+# Arguments
+
+- `reducer`: any function that can consume an iterable as its only argument, such as `sum` or `Statistics.mean`.  Can also be `identity` or any other user-defined function!
+- `operator`: any function that can consume a named tuple as its only argument, such as `mean` or `std`.  Applied on the set of values from RasterStacks, what you would get when you index into a RasterStack like `rasterstack[1, 1]`.
+- `data`: a `RasterStack`
+- `of`: a `DimTuple`, `Extent`, $OBJ_ARGUMENT
+
+# Keywords
+$GEOMETRYCOLUMN_KEYWORD
+These can be used when `of` is or contains (a) GeoInterface.jl compatible object(s):
+
+- `shape`: Force `data` to be treated as `:polygon`, `:line` or `:point`, where possible.
+- `boundary`: for polygons, include pixels where the `:center` is inside the polygon,
+    where the line `:touches` the pixel, or that are completely `:inside` inside the polygon.
+    The default is `:center`.
+- `progress`: show a progress bar, `true` by default, `false` to hide..
+- `skipmissing`: wether to apply `f` to the result of `skipmissing(A)` or not. If `true`
+    `f` will be passed an iterator over the values, which loses all spatial information.
+    if `false` `f` will be passed a masked `Raster` or `RasterStack`, and will be responsible
+    for handling missing values itself. The default value is `true`.
+
+# Example
+
+"""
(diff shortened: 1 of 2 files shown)


src/methods/mapzonal.jl
```diff
@@ -0,0 +1,130 @@
...
+
+"""
+function mapzonal(reducer, operator, data::RasterStackOrArray; of, kw...)
+    _mapzonal(reducer, operator, data, of; kw...)
+end
+
+function _mapzonal(reducer, operator, x, ext::Extents.Extent; skipmissing = true)
+    cropped = crop(x; to = ext, touches = true)
+    prod(size(cropped)) > 0 || return missing
+    # We can't use skipmissing here, since it doesn't work on rasterstacks
```
rafaqz, line 38: We should just make this work, but it's just not totally clear what the rule is - any missing or all missing

rafaqz (commented):
Can we not do this already with `mapslices` when `skipmissing=false`? 

Or is it a performance issue

src/methods/mapzonal.jl
asinghvi17: In here I've implemented the "all missing" variant but it's up for debate

asinghvi17: I guess the question is, does `skipmissing` mean "skip all values that are missing in general" or "skip all values that are missing because of my mask"?  The first is the most useful because there can be functions that don't tolerate missings

asinghvi17:
To do this with `mapslices` you would have to essentially remake `zonal` anyway (crop + mask and walk down the tree of geometry), but I can check the performance of that

rafaqz:
I mean inside the zonal function, can you just call mapslices on the RasteStack it gives you?

asinghvi17:
It doesn't give you a rasterstack though, since it iterates over layers...

Unless you mean changing the implementation?  That would be breaking I think

rafaqz:
Hmm yes that's annoying. How do we get the best of both behaviours...

But can we workshop how to get this all into `zonal` elegantly? I'm happy to break things at this stage, or at least add options rather than more functions

asinghvi17:
Hmm, it could be a switch that does this internally?

rafaqz:
`bylayer=true` ?

asinghvi17:
There are two scenarios here - "set of rasters" (where you return multiple columns) or "linked rasters" (where you want a slice at those x and y coordinates)

asinghvi17:
The thing is that methods won't necessarily work the same in both, or work at all...

asinghvi17:
Hmm, with context `bylayer` does make sense

rafaqz:
But if we have `bylayer=true, skipmissing=false` and get passed a complete masked `RasterStack` in `f`, will that let you do what you need with `mapslices`?

rafaqz:
>How  would you treat 3d rasters in zonal then? 

With `skipmissing=false` you already get the whole object, whatever it is. Otherwise it's a  `skipmissing` iterator over all dimensions so `mean` will just work. 

With `bylayer=false` stacks can be the same, and you handle slicing however you need to as with any `DimStack`

asinghvi17:
Succeeded by #829

PR #776, opened 2024-10-02
Print a styled backend error message so it's more eye-catching

asinghvi17:
Screenshot 2024-10-02 at 2 09 57 PM

### DIFF
--- a/src/sources/sources.jl
+++ b/src/sources/sources.jl
@@ -59,7 +59,7 @@ end
 # error message to show when backend is not loaded
 function Base.showerror(io::IO, e::BackendException)
     printstyled(io, "Rasters.jl"; underline = true)
-    printstyled(io, " requires backends to be loaded externally as of version 0.8.  Run ")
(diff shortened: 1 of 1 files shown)


rafaqz:
True somehow people just don't see that bit

src/sources/sources.jl
```diff
@@ -58,7 +58,10 @@ end
 
 # error message to show when backend is not loaded
 function Base.showerror(io::IO, e::BackendException)
-    print(io, "`Rasters.jl` requires backends to be loaded externally as of version 0.8. Run `import $(e.backend)` to fix this error.")
+    printstyled(io, "Rasters.jl"; underline = true)
+    printstyled(io, " requires backends to be loaded externally as of version 0.8.  Run ")
```
rafaqz, line 62:
```suggestion
    printstyled(io, " requires backends to be loaded manually.  Run ")
```
I think we can cut the version 0.8 part now

Issue #774, opened 2024-10-02
Potential package compatibility issue (Rasters.jl, GeoMakie.jl)


nicholasbalasus:
Starting with Julia version 1.10.5, I make a new environment and add `Rasters` and then `GeoMakie`. Notably, when I add `GeoMakie`, I get an `✗` next to Rasters.

```
Precompiling project...
  ✗ Rasters → RastersMakieExt
  1 dependency successfully precompiled in 30 seconds. 259 already precompiled.
```

Then, in the REPL, I run `using Rasters` and then `using GeoMakie` which gives a long error.
```
julia> using Rasters

julia> using GeoMakie
Precompiling RastersMakieExt
        Info Given RastersMakieExt was explicitly requested, output will be shown live 
ERROR: LoadError: ArgumentError: invalid type for argument t in method definition for convert_arguments at /n/home06/nbalasus/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/plotrecipes.jl:308
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/plotrecipes.jl:308
 [2] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [3] include(x::String)
   @ RastersMakieExt ~/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/RastersMakieExt.jl:1
 [4] top-level scope
   @ ~/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/RastersMakieExt.jl:12
 [5] include
   @ ./Base.jl:495 [inlined]
 [6] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2222
 [7] top-level scope
   @ stdin:3
in expression starting at /n/home06/nbalasus/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/plotrecipes.jl:308
in expression starting at /n/home06/nbalasus/.julia/packages/Rasters/0kAkQ/ext/RastersMakieExt/RastersMakieExt.jl:1
in expression starting at stdin:3
  ✗ Rasters → RastersMakieExt
  0 dependencies successfully precompiled in 13 seconds. 239 already precompiled.
[...]

PR #777, opened 2024-10-02
Remove discretesurface dispatch

asinghvi17:
Fix https://github.com/rafaqz/Rasters.jl/issues/774


asinghvi17:
The problem is that DDMakieExt will still fail because it probably has this issue too..

PR #723, opened 2024-09-02
Continue patch

lazarusA:
continuation of https://github.com/rafaqz/Rasters.jl/pull/722

PR #780, opened 2024-10-07
add continue_patch changes more cleanly: fixes Makie for new DD

rafaqz:
supercedes #723 

It was too hard to fix the commit history there!

@lazarusA this should have everything from the other one not already in main

I'll credit you in the quash commit


lazarusA:
All good 👍.  I also noticed that, but no time to fix. I was also planning,  eventually,  to start from scratch again.

rafaqz:
Ah looks like DD has broken something too

lazarusA:
for docs, is still not getting the latest release of DD. Maybe delete the `cache` for this PR?

rafaqz:
Ah right how do we actually do that again?

lazarusA:
wow! DD is really breaking things up. Next release will be definitely breaking :D

rafaqz:
The equality ones I had earlier will mostly just affect tests and you can turn the off with `strict=false` in `@d` 

But this one now is allocations I. `write`, like type instability? But I can reproduce it at all

tiemvanderdeure:
We added those allocation checks because the eachchunk didn't work and the values were written one by one, which led to millions of allocations. The test only exceeds whatever arbitrary boundary I set by a little bit and only in one place. So while it sure is weird maybe it's okay to just increase it so tests pass, as long as we are sure there is not some serious performance issue.

tiemvanderdeure:
I realized this could be due to julia 1.11?

rafaqz:
Yeah I think so too... Have to update and check type stability

PR #783, opened 2024-10-09
add a constructor to load future worldclim as a rasterstack

tiemvanderdeure:
Future bioclim data provided by worldclim is a single .tif file, whereas current comes as a single file for each bioclim variable.

This PR just adds a constructor that makes a RasterStack for future bioclim that matches the one you would get from RasterStack(WorldClim{Bioclim})

This will be relevant whenever https://github.com/EcoJulia/RasterDataSources.jl/pull/69 is merged.


rafaqz:
A test would show that its currently broken ;)

rafaqz:
Also RDS is merged and bumped to 0.7.0 now so this can happen for real

tiemvanderdeure:
Do we want to make 0.7.0 the minimum, or just add it? I know it was super breaking for these future datasets but the interfacing with rasters didn't change at all and these changes also don't require RDS 0.7.0

rafaqz:
Oh nice if that didn't break anything we can keep both

tiemvanderdeure:
They need 0.7 to pass the tests, but they didn't break anything earlier. We just made constructors for dispatches that were added in 0.7.

The rasterdatasources tests don't actually run in the CI, by the way.

rafaqz:
Oh yeah we turned them off because of ucdavis.

What I mean to say is we can only include 0.6 if those tests pass on 0.6, otherwise force 0.7

tiemvanderdeure:
the old tests will still pass, but the ones added in this PR won't. So I'll limit to 0.7.

Project.toml
```diff
@@ -70,7 +70,7 @@ NCDatasets = "0.13, 0.14"
 OffsetArrays = "1"
 ProgressMeter = "1"
 Proj = "1.7.2"
-RasterDataSources = "0.5.7, 0.6"
+RasterDataSources = "0.5.7, 0.6, 0.7"
```
rafaqz, line 73:
```suggestion
RasterDataSources = "0.7"
```

tiemvanderdeure:
Also just fixed a test that has been broken by https://github.com/rafaqz/Rasters.jl/pull/613 but we never noticed.

PR #790, opened 2024-10-11
do not unnecessarily return a union in extract

tiemvanderdeure:
when `extract`ing with skipmissing=false, the geometry type will currently always be a `Union{Missing, ...}`, even if none of the provided geometries were missing.

This PR makes it so that the type of the geometries in input and output will match.

I had to change `_rowtype` to do this, and changed it a little more to make it easier to reuse (which I want to do for Rasters.sample). 

I also added some missing text to the docstring.

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -5,7 +5,7 @@ istrue(::_True) = true
 istrue(::_False) = false
 
 """
-    extract(x, data; atol)
+    extract(x, data; [geometry, index, name, skipmissing, atol])
 
 Extracts the value of `Raster` or `RasterStack` at given points, returning
 an iterable of `NamedTuple` with properties for `:geometry` and raster or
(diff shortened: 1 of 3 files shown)


tiemvanderdeure:
Not 100% sure if this is a bugfix or a slightly breaking change, though.

src/methods/extract.jl
```diff
@@ -5,7 +5,7 @@ istrue(::_True) = true
 istrue(::_False) = false
 
 """
-    extract(x, data; atol)
+    extract(x, data; [geometry, index, name, skipmissing, atol])
```
rafaqz, line 8:
```suggestion
    extract(x, data; kw...)
```

At this point people can look at the Keywords section

src/methods/extract.jl
```diff
@@ -311,50 +311,44 @@ end
...
+        typeof(g)
+    end
+    _rowtype(x, G; geometry, index, skipmissing, names)
 end
-function _rowtype(x, ::Type{G}, geometry::_True, index::_False, skipmissing::_True, names::NamedTuple{Names}) where {G,Names}
-    keys = (:geometry, Names...,)
-    types = Tuple{G,_nametypes(x, names, skipmissing)...}
+function _rowtype(x, g::Type; geometry, index, skipmissing, names, kw...)
+    keys = _rowkeys(geometry, index, names)
+    types = _rowtypes(x, g, geometry, index, skipmissing, names)
```
rafaqz, line 324:
```suggestion
function _rowtype(x, ::Type{G}; geometry, index, skipmissing, names, kw...)
    keys = _rowkeys(geometry, index, names)
    types = _rowtypes(x, G, geometry, index, skipmissing, names)
```

Don't give julia the option of not propagating G as a known type

tiemvanderdeure:
Just copied the `_rowtype` changes to this branch so we can figure out type stability in this PR and how Rasters.sample should work in the other.

rafaqz:
3M allocations to 13 is pretty nice!

PR #791, opened 2024-10-11
Add `Rasters.sample`

### LINKED ISSUE #771: Add random sampling functionality (like R's `terra::spatSample()`)
Thanks for a great package! This is a feature request for a function like R’s `terra::spatSample()` that enables random sampling of locations on a raster, optionally excluding NaN/missing values. Below is some quick-and-dirty example code which demonstrates the functionality, but is inefficient when NaNs should be ignored. Maybe this is already quick & easy to do in Julia? (I am coming from R.)  

```julia
# Convert Raster to DataFrame, optionally dropping NaNs 
function asDataFrame(; x::Rasters.Raster, na_rm::Bool = true)
    # Get DataFrame
    xyz = DataFrame(x)
    rename!(xyz, ["x", "y", "z"])
    # Filter non NA cells 
    if na_rm
        xyz = xyz[.!isnan.(xyz.z), :]
    end
    return xyz
end 

# terra::spatSample() replacement, via asDataFrame()
function spatSample(; x::Rasters.Raster, size::Int64 = 1, na_rm::Bool = true)
    xyz = asDataFrame(x = x, na_rm = na_rm)
    xyz[sample(1:nrow(xyz), size, replace = true), :]
end 
```

tiemvanderdeure:
This PR adds a `Rasters.sample` function in a StatsBase extension, as suggested here: https://github.com/rafaqz/Rasters.jl/issues/771

The keywords correspond to `extract` keywords. Some of the internals are also re-used.

I would suggest not exporting this to avoid confusion with StatsBase.sample, which has a totally different API.

If `skipmissing = true`, Rasters.sample and extract will always return the same types - otherwise it depends on the element type of the Raster. I tried to avoid creating `Union{Missing, ...}` as much as possible. 

Closes #771


tiemvanderdeure:
Unfortunately the conversion step still makes this function with skipmissing = true slow for Rasters that have Union types. This happens inside `_maybe_add_fields`, which is an internal from `extract`, so I don't know if it's possible to get rid of this or how.

Just something like
```julia
ga = Raster(A, (X(1.0:1:2.0), Y(1.0:1:2.0)); missingval=missing) 
Rasters.sample(ga, 1_000_000, skipmissing = true)
```
takes 0.3 seconds because of this:
image

tiemvanderdeure:
Tests fail because of `[5453] signal 11 (128): Segmentation fault`. What a joy.

ext/RastersStatsBaseExt/sample.jl
```diff
@@ -0,0 +1,92 @@
...
+    StatsBase.sample(rng, CartesianIndices(x), wts, n; replace, ordered)
+end
+function sample_indices(rng, x, n, skipmissing::_True, weights::AbstractDimArray, replace, ordered, weightstype)
+    wts = weightstype(vec(@d boolmask(x) .* weights))
+    StatsBase.sample(rng, CartesianIndices(x), wts, n; replace, ordered)
+end
+
+# Determine the row type, making use of some of extract machinery
+_srowtype(x, g; kw...) = _srowtype(x, typeof(g); kw...)
+function _srowtype(x, g::Type; geometry, index, skipmissing, names, kw...)
```
rafaqz, line 66: This seems like a lot of duplication with `extract`... can we generalise them to pass in `_rowtypes` or `_srowtype`?

Then they can go in `utils.jl`

ext/RastersStatsBaseExt/sample.jl
```diff
@@ -0,0 +1,92 @@
+Rasters.sample(x::RA.RasterStackOrArray, n::Integer; kw...) = Rasters.sample(Random.GLOBAL_RNG, x, n; kw...)
+@inline function Rasters.sample(
```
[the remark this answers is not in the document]
tiemvanderdeure: I think we talked about it and I even wrote it in the docstring. We call StatsBase.sample under the hood, but we add a whole lot of keywords and totally change the output, so I think it would be confusing to add it as a method of StatsBase.sample. Plus `StatsBase.sample(x::Raster)` already works and could also still be useful after adding this.

ext/RastersStatsBaseExt/sample.jl
tiemvanderdeure: I tried my best here so at least there's less code in sample, though it's still probably more convoluted than it needs to be

rafaqz: It just seems like there is a structural similarity that we can move to a single more abstract function. But I may be wrong

tiemvanderdeure: I changed this last night!

tiemvanderdeure:
Sorry for all the mess with the rebases. I added an extra keyword to `_rowtype` to distinguish between whether index/geometry fields may be missing, and wether values extracted from the raster may be missing. The former is never the case in `sample`. So now we can completely reuse the same internal.

tiemvanderdeure:
This is still not type stable and I can't quite figure out why...

rafaqz:
> Somehow NamedTuple(x[idx])[K] is not type stable and I can't quite figure out why...

Was going to say that looks dubious. Probably we should subset the stack with `st[K]` before we get to this method to reduce indexing. Then you can just do:

`x[idx]`

tiemvanderdeure:
> Was going to say that looks dubious. Probably we should subset the stack with `st[K]` before we get to this method to reduce indexing. Then you can just do:
> 
> `x[idx]`

I tried this but it failed tests, because `st[K]` does not necessarily have all the dimensions of `st`...

tiemvanderdeure:
This should be type stable now in all cases.

I would still like to change the `geometry` keywords, which now is not very flexible. It could be nice to allow users to specify the type of the geometry column, and which dimensions should be used. So something like `geometry = (Y=Y, X=X)` would return a Y-X NamedTuple with data from the Y and X dimensions.

tiemvanderdeure:
Tests are still failing on a segmentation fault in resample 😭

rafaqz:
Ahh that's interesting about the dimensions. I guess we need to be consistent and return a table of the same length every time.

You can put the `resample` tests last to get around them for now. But we will need to deal with them before bumping a version.

rafaqz:
We can totally let users specify the geometry like that too.

tiemvanderdeure:
> Ahh that's interesting about the dimensions. I guess we need to be consistent and return a table of the same length every time.

It's how we designed `extract` so I think it's good to be consistent. `extract(stack, point; names)` is not the same as `extract(stack[names], points)`.

rafaqz:
Maybe we should put that exact line in the `extract` docs

ext/RastersStatsBaseExt/sample.jl
```diff
@@ -0,0 +1,89 @@
...
+function _getindices(::Type{T}, x, points, indices) where T
+    rows = Vector{T}(undef, size(indices))
+    for (i, I) in enumerate(indices)
+        rows[i] = _getindex(T, x, points, I)
+    end
+    return rows
+end
+
+_getindex(::Type{T}, x::AbstractRasterStack{<:Any, NT}, points, idx) where {T, NT} = 
+    RA._maybe_add_fields(T, NT(x[RA.commondims(idx, x)]), points[RA.commondims(idx, points)], val(idx))
```
tiemvanderdeure, line 52: Explicitly using the eltype of the stack like this was the only way I could find in which this would compile away. But it looks somehow ugly and I don't know if there's a better way?

tiemvanderdeure:
> Maybe we should put that exact line in the `extract` docs

Maybe. This feature justifies the having the names keyword in the first place, but it probably has a limited number of use cases.

tiemvanderdeure:
I added the option to specify a geometry type, even though it now only works for Tuple and NamedTuple. For other types we can't just call `convert`, so it's not so easy to make it more generic.

ext/RastersStatsBaseExt/sample.jl
rafaqz: Like what? Is this comment on the wrong line?

ext/RastersStatsBaseExt/sample.jl
```diff
@@ -0,0 +1,85 @@
...
+    T = RA._rowtype(x, geometrytype; geometry, index, skipmissing, skipinvalid = _True(), names)
+    x2 = x isa AbstractRasterStack ? x[K] : RasterStack(NamedTuple{K}((x,)))
+    _getindices(T, x2, points, indices)
+end
+
+_getindices(::Type{T}, x, points, indices) where T = 
+    broadcast(I -> _getindex(T, x, points, I), indices)
+
+_getindex(::Type{T}, x::AbstractRasterStack{<:Any, NT}, points, idx) where {T, NT} = 
+    RA._maybe_add_fields(T, NT(x[RA.commondims(idx, x)]), points[RA.commondims(idx, points)], val(idx))
```
rafaqz, line 48: It's weird this NT trick is needed, maybe DD stack getindex stability is broken on 1.11?

Looks like something needs to be forced somewhere

rafaqz:
Is there a test for the NamedTuple geometry?

ext/RastersStatsBaseExt/sample.jl
tiemvanderdeure: I know, nothing else worked. Even `AbstractRasterStack{K}` and then `NamedTuple{K}` wasn't type stable.

tiemvanderdeure:
> Is there a test for the NamedTuple geometry?

Yes, although it's pretty rudimentary.
```
    @test typeof(first(
        Rasters.sample(StableRNG(123), test, 2, geometry = (X = X, Y = Y))
    ).geometry) <: NamedTuple{(:X, :Y)}
```

ext/RastersStatsBaseExt/sample.jl
rafaqz: No that makes sense, it's the eltype that's not propagating.

We might need some more `Base.assume_effects` in DD indexing

rafaqz:
Ok that's fine, I like the syntax of just passing the tuple. Does it work for symbols and constructed dims too?

tiemvanderdeure:
It just calls `commondims`, so I think constructed dims yes, symbols no. This is parsed and a ` _True()` or `_False()`, a dimtuple, and a type are passed to `_sample`, so that should all be type stable. 

NamedTuple types still allocate though, idk if/how to avoid that?

rafaqz:
That's what I mean. For NamedTuple you need a NamedTuple of constructed dimensions, not types or symbols. That can be type stable making NamedTuple points, the others can't be

src/methods/extract.jl
```diff
@@ -285,27 +277,8 @@ Base.@assume_effects :total function _maybe_add_fields(::Type{T}, props::NamedTu
...
-    _extractrowtype(x, g, I; geometry, index, skipmissing, names, kw...)
-end
-function _extractrowtype(x, g, ::Type{I}; geometry, index, skipmissing, names, kw...) where I
-    G = if istrue(skipmissing)
-        nonmissingtype(typeof(g))
-    else
-        typeof(g)
-    end
-    _extractrowtype(x, G, I; geometry, index, skipmissing, names, kw...)
+    end |> T
```
tiemvanderdeure, line 280: Apparently this `|> T` was needed to avoid allocations when geometry is a namedtuple.

asinghvi17: Coyld try ::T instead of piping too, it's a bit clearer

tiemvanderdeure: It already had a ::T, but it was still allocating in cases like https://github.com/rafaqz/Rasters.jl/pull/790#issuecomment-2408503919 and also https://github.com/rafaqz/Rasters.jl/pull/791#issuecomment-2411420187, though I'm not sure why. I think it's something with the compiler giving up on the nested namedtuple structures?

rafaqz: Yeah I thought those were functionally the same thing. ::T should call `convert`, but maybe the constructor is better.

The stability is the compiler knowing anything unstable is also an error, maybe that's just clearer with |> for some reaon

tiemvanderdeure:
I just added a dispatch so `Rasters.sample(myraster)` also works and returns just a NamedTuple. I don't have anything else to add here. The methods tests are before resample so if it segfaults again then tests have passed.

rafaqz:
Amazing let's merge

rafaqz:
Ohh maybe some links in the methods page of the docs would have helped

tiemvanderdeure:
True, I forgot about that

PR #799, opened 2024-10-11
Add tests for unitful `cellarea`

asinghvi17:
Does what it says on the tin.  Unfortunately it looks like Unitful can't `convert` to Float64 (see https://github.com/PainterQubits/Unitful.jl/issues/742) so going to Proj is where this fails.

It also occurs to me that we could try to promote and rebuild e.g. bignumbers or Unitful quantities when going into `reproject`.  That would allow unitful axes to pass through reproject unchanged.

### DIFF
--- a/test/cellarea.jl
+++ b/test/cellarea.jl
@@ -1,6 +1,7 @@
 using Rasters, DimensionalData, Rasters.Lookups, Proj
 using Test
 using DimensionalData: @dim, YDim
+import Unitful
 include(joinpath(dirname(pathof(Rasters)), "../test/test_utils.jl"))
 
 @testset "cellarea" begin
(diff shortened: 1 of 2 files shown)


rafaqz:
Hmm with units, probably we need to check they are the same units as the crs and otherwise convert. You can ` ustrip` to a specific unit, like `ustip("m", x)`

asinghvi17:
That sounds like a way bigger effort :D, but doable at some point via Proj.  It might be nice for reproject, actually, since we could discover the CRS's units (maybe stash that in lookup metadata) and this would all optimize away at compile-time.  Then indexing with arbitrary units could just work TM via Unitful's automatic conversion.

test/cellarea.jl
```diff
@@ -70,4 +71,29 @@ include(joinpath(dirname(pathof(Rasters)), "../test/test_utils.jl"))
...
+         unitful_dimz_3857 = (ux_3857, uy_3857)
+         @test cellarea(Planar(), unitful_dimz_3857) isa Raster{<:Unitful.Quantity}
+         @test Unitful.ustrip.(parent(cellarea(Planar(), unitful_dimz_3857))) == parent(cellarea(Planar(), dimz_3857))
+         # Unitful lookups shouldn't work without a unitful manifold
+         @test_throws Unitful.DimensionError cellarea(unitful_dimz_3857)
+         # The tests below fail because Unitful apparently can't convert to Float64...
+         # (see https://github.com/PainterQubits/Unitful.jl/issues/742)
+         # But we also have to re-convert back to the original unit type, which cellarea currently
+         # doesn't do.
+
```
rafaqz, line 94:
```suggestion
         # This needs fixes for unit handling

```

Unitful is like a free test suite for physical models not converting is by design!

If we are going to make units work here we need to do `ustrip(u"m", x)` to ensure we are actually getting metres coming in. If people pass in kilometers and we convert back and forth with Proj.jl treating them as meters in the middle, we are really breaking the contract.

asinghvi17: That makes sense.  Maybe we have a `maybe_ustrip(unit, x)` defined in Rasters, its definition for Unitful `unit` in `RastersUnitfulExt`, and its definition for Proj units in `RastersProjUnitfulExt`?

(This is getting to be a mouthful..)

test/cellarea.jl
```diff
@@ -70,4 +71,29 @@ include(joinpath(dirname(pathof(Rasters)), "../test/test_utils.jl"))
...
+         # Case 1: cellarea with unitful manifold
+         # This is the simplest case
+         unitful_manifold = Spherical(; radius = Spherical().radius * Unitful.u"m")
+         unitful_cellarea = cellarea(unitful_manifold, dimz_3857)
+         @test unitful_cellarea isa Raster{<:Unitful.Quantity}
+         @test Unitful.ustrip.(parent(unitful_cellarea)) == cellarea(Spherical(; radius = unitful_manifold.radius |> Unitful.ustrip), dimz_3857)
+         # Case 2: unitful dimensions
+         ux_3857 = rebuild(x_3857; val = rebuild(val(x_3857); data = val(x_3857) .* Unitful.u"m", span = Regular(val(x_3857).span.step * Unitful.u"m")))
+         uy_3857 = rebuild(y_3857; val = rebuild(val(y_3857); data = val(y_3857) .* Unitful.u"m", span = Regular(val(y_3857).span.step * Unitful.u"m")))
+         unitful_dimz_3857 = (ux_3857, uy_3857)
```
asinghvi17, line 85: This is also super complicated, should we have DD overloads to make constructing lookups with unitful easier?

test/cellarea.jl
rafaqz: Yeah those names! But makes sense. Somehow we need some mapping from Proj units to Unitful.

felixcremer: Why does conversion from Proj to Unitful have to live in Rasters.jl? Shouldn't this be in a ProjUnitfulExt?

asinghvi17: Hmm, we could introduce a nice interface to batch transform an array, but do it out of place in Proj.jl

asinghvi17: I was initially going to put that here because this calls the C-API directly, but if it doesn't have to then there's no reason to stop this from calling Proj.

test/cellarea.jl
rafaqz: I think `set` will just work this out for use, not need to `rebuild` manually?

asinghvi17: We need apply for lookups to get at the base data, otherwise it won't work generically...

test/cellarea.jl
rafaqz: I think we need a more general package for parsing geospatial units to Unitful. There is Proj, gdal, CF conventions, and some overlap between them.

Maybe UnitfulGeo.jl and it can go in JuliaGeo

test/cellarea.jl
rafaqz: Yeah we need something like `modify` but elementwise on lookups. 

This could work?
```julia
set(A, X => v -> u"m" * v) 
```

We would need to check the result is still regular and error if not.

Issue #797, opened 2024-10-11
`missing` is not materialized when elementwise dividing two rasters?


asinghvi17:
MWE:
```julia
using Rasters, RasterDataSources, ArchGDAL
precip = Raster(WorldClim{Climate}, :prec; month = 6)
areas = cellarea(precip)
precip ./ areas
```
 
![download-32](https://github.com/user-attachments/assets/577304e6-b9db-4d40-a583-39385246a6e1)

but the precip dataset has a bunch of missings.  I'm not really sure how this happened.

asinghvi17:
The workaround I used was to mask both the rasters by the same geometry.  But that's not generally applicable.

rafaqz:
You need to `replace_missing` first until the `cf` PR is merged

asinghvi17:
Ah perfect, if that fixes it then I'll just leave the code in for now.

PR #789, opened 2024-10-11
[WIP] More docs

asinghvi17:
- Switch all references for reproject from GDAL to Proj, improve docstring
- Add Proj to the docs project
- Slightly clarify some doc examples

TODOs:
- [ ] Add official tutorials section with Literate examples
- [ ] Add detailed doc page and exploration of cellarea
- [ ] Add docs about the many ways to change the CRS of a raster
- [ ] Merge in Alex's read/write docs PR here
- [ ] Merge in Lazaro's https://github.com/rafaqz/Rasters.jl/pull/730/ resample PR here.

PR #800, opened 2024-10-13
Add cellarea docs, allow Literate.jl tutorials in the doc pipeline, fix typos

asinghvi17:
- Switch all references for reproject from GDAL to Proj, improve docstring
- Add Proj to the docs project
- Slightly clarify some doc examples
- Add Literate.jl example pipeline
- Add detailed doc page and exploration of `cellarea`


asinghvi17:
Supersedes #789

src/methods/reproject.jl
```diff
@@ -2,15 +2,16 @@
 """
     reproject(obj; crs)
 
-Reproject the lookups of `obj` to a different crs. 
+Reproject the lookups (axes) of `obj` to a different crs.
```
rafaqz, line 5: We try to maintain the distinction between axes and lookups as different things

rafaqz (commented):
My main thought here is why doesn't the `cellarea` tutorial just go in the `cellarea` API docs?

Or in extended docs? This approach feels like a lot of text needs to be repeated and the overall complexity of the documentation is increased. 

To me tutorials should be more about combining multiple methods into workflows.

lazarusA:
your cellarea tutorial is not running the lines as intended (no output in most). Plus, is hard to find, I needed to look for it in the search box, maybe add a new section to the nav bar?

lazarusA:
In yax, we have something like, although is not very much maintained 🫨 .

Screenshot 2024-10-13 at 15 44 46

lazarusA:
@rafaqz IMHO, API docs should not have extended docs, and be as clean as possible to the main description of input and output arguments.  For developers it might be easier to see things there if they work as intended, another test, but documenting should have a low entry level, specially for people (me) that don't know how to use those functions 😄 .

rafaqz:
Sure, I just meant using extended so you can add more examples for the online docs that don't swamp the repl. Not so you would actually look for them in the repl.

My main point is not to have `cellarea` docs twice,  I don't want too much duplicate text to maintain.

src/methods/reproject.jl
asinghvi17: I see what you're saying, the problem is that even people who've used rasters for a while have no clue what "lookups" are.  There has to be some frame of reference.  "Dimension values" is also super unclear - what does that mean?

Maybe we can rewrite the parentheses to be more clear?  Would "axis values" work there?

asinghvi17: FWIW, this is meant to be for people who have no clue what cellarea is or why you might want it.  So you have to go from the ground up, more or less.

asinghvi17:
Even extended docs are a little button you can click on, I don't see an alternative to tutorials unfortunately.  These also serve very different uses since the tutorials are to introduce someone to a function and what you can do with it, REPL docs are to remind you of what the function itself does and how to call it

src/methods/reproject.jl
rafaqz: I get it, but we just can't use a common base function name that also works on a Raster to explain a DD function that's really a different thing.

It's clearly difficult to find words that describe lookups without semantic overloads (see AxisKeys.jl... keys are a different base method again) but it's important that we try. 

I think axis values is helpful, but maybe values is also a pretty empty word. Maybe this needs to be systematic from DD up, we could workshop the language over there.

rafaqz:
rasterio also just uses the api docs for functions like this
https://rasterio.readthedocs.io/en/stable/cli.html#rasterize

And their tutorials are reserved for general things

From my perspective, contributors here and for DD have come and gone, and in some years there is a good chance I will have to maintain these docs. Minimizing that task across what are now very many packages is becoming central to maintaining my sanity.

If you make tutorials for some Rasters, methods, at some point we are committing to making and maintaining them for all methods. 

I _really_ want to try putting these tutorials in the regular API docs, and changing that when they are really to big and we are committed to fleshing out the docs for every other method here.

rafaqz:
Currently we have methods listed here:
https://rafaqz.github.io/Rasters.jl/stable/methods

They could be better organised in a menu somehow. But to me just improving those is better than adding additional versions of them all. Like `cellarea` isn't even listed there yet.

asinghvi17:
My issue with putting these in docstrings is twofold:
- They don't _run_, so they can't be used as additional sanity checks.  Output may change over time and the REPL examples would stay static.  This is another source of maintenance burden.
- The kind of examples I want to show (which are, fundamentally, full-workflow examples) are unsuitable to REPL examples because they aren't _visual_ enough.  In the REPL, I have no way to show a raster's values without completely overwhelming the user with information.  A plot, on the other hand, is simple and immediately understandable. 

  Looking at the rasterio docs you sent - when will I ever actually run those commands?  For that matter, when will a user ever run the example commands in the REPL docs, if they don't know they are there?  This is the missing source of information I want to correct here.  People don't fundamentally care unless things are immediately in front of them, so they can see what it does and see the benefit.

That being said, I can see your point about implicit promises and maintaining stuff.  How about this:
- Move the cellarea tutorial to a new "Estimating spatial means" tutorial (and keeping the tutorial structure)
- Move some of the explanatory text here to the cellarea docs

asinghvi17:
If the goal of the geocompx books is to replicate the R and Python versions directly then I'm not sure if we can sneak these examples in there, though there will definitely be more examples of these kinds of functions.

rafaqz:
The Rasters API docs already have examples with plots! And they do actually run :)

I like the split - estimating spatial means is a complete workflow task, perfect for a tutorial

rafaqz:
I actually think all of the current API docs can be better and longer with more explanation and examples, I just haven't had time

docs/src/tutorials/methods/cellarea.jl
```diff
@@ -0,0 +1,114 @@
...
+!!! note
+    If you made it this far, congratulations!  
+
+    It's interesting to note that we've replicated the workflow of `zonal` here.  
+    `zonal` is a more general function that can be used to compute any function over geometries, 
+    and it has multithreading built in.  
+    
+    But fundamentally, this is all that `zonal` is doing under the hood - 
+    masking and cropping the raster to the geometry, and then computing the statistic.
+=#
```
lazarusA, line 114: from experience (a painful one), I can tell you that for long explanations using literate and `.jl` files is not ideal. I did that for a while in the yax docs, and eventually I went back to just markdown. For small scripts that just run is fine, like in beautifulmakie, but once you start doing bigger things I would say that markdown should be the way to go.

asinghvi17: Hmm, why not?  It's worked pretty well for me so far, but those were all in situations where my markdown content doesn't exceed ~150 lines.  Curious what your issues were...

asinghvi17: I do at some point want to integrate a Quarto like system here, since that seems to be the best at this stuff.

rafaqz: I agree with @lazarusA, Literate.jl adds a layer of complexity that's just not worth it in some cases. 

I would prefer not to have it here

src/methods/reproject.jl
asinghvi17: Maybe we could say "x and y values" for now?

rafaqz: Lookup values ?

asinghvi17: Then we're back to the problem of "what is a lookup" again?

docs/src/tutorials/methods/cellarea.jl
```diff
@@ -0,0 +1,114 @@
...
+
+using Rasters
+import Proj # to activate the spherical `cellarea` method
+import ArchGDAL # purely for data loading
+
+# To construct a raster, we'll need to specify the x and y dimensions.  These are called "lookups" in Rasters.jl.
+using Rasters.Lookups
+# We can now construct the x and y lookups.  Here we'll use a start-at-one, step-by-five grid.
+# Note that we're specifying that the "sampling", i.e., what the coordinates actually mean,
```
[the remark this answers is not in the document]
asinghvi17: I might take this section out, not sure yet.

docs/src/tutorials/methods/cellarea.jl
```diff
@@ -0,0 +1,114 @@
...
+cropped_precip = crop(precip; to = chile)
+# Now, we mask the data such that any data outside the geometry is set to `missing`.
+masked_precip = mask(cropped_precip; with = chile)
+heatmap(masked_precip)
+# This is a lot of missing data, but that's mainly because the Chile geometry we have encompasses the Easter Islands as well, in the middle of the Pacific.
+
+# Now, let's compute the average precipitation per square meter across Chile.
+# First, we need to get the area of each cell in square meters.  We'll use the spherical method, since we're working with a geographic coordinate system.  This is the default.
+areas = cellarea(masked_precip)
+masked_areas = mask(areas; with = chile)
```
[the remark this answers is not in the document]
asinghvi17: https://github.com/rafaqz/Rasters.jl/issues/797

src/methods/reproject.jl
felixcremer: Maybe we could say "location information".  But I would prefer to just use lookups here and have a glossary where we would explain these differences, because otherwise we would have to duplicate this info everywhere we use lookups.

asinghvi17: I would argue that we _should_ duplicate this information everywhere we use lookups, at least in top-level, user accessible functions.  I don't want to overestimate the curiosity of a new user v/s their unwillingness to click a link.

asinghvi17:
hold on, I just realized that these examples don't actually run!  The code block language is julia and not julia-doctest.  And they don't get rendered in the REPL, either.


I wonder if it makes sense to dump the contents of the docstring directly into an auto-generated page?  And then maybe have a switch to replace julia with `@example` or something.   This way, the doc examples can be static, but still run in the docs for output.

src/methods/reproject.jl
asinghvi17: That's not to say that we shouldn't have a glossary with a more detailed explanation, but there should be something like:

[lookup values](link to DD lookup docs) (the positions of the cells in each dimension)

or something

asinghvi17: Actually "lookup values (indicating the positions of each cell)" sounds pretty descriptive

asinghvi17: Actually "lookup values (indicating the positions of each cell)" sounds pretty descriptive

asinghvi17:
@rafaqz now I remember why I was against jldoctest blocks - they don't show up in the REPL at all!  So they're fundamentally useless for in-REPL documentation.

asinghvi17:
Ah I guess we're missing the `julia>` prompt (that's a super weird thing to require IMO)

https://github.com/KristofferC/OhMyREPL.jl/issues/293#issuecomment-1459920171

rafaqz:
Yeah for me e.g. `rasterize` has a full code example

lazarusA:
incorporated in #844

PR #801, opened 2024-10-13
[WIP] resample/warp tutorial

asinghvi17:
Adds a tutorial about resample/warp to the docs.


rafaqz:
Again my question is if this should be in the API docs?

Better docs for all these methods are really needed, but keeping maintenance efficient is really important long term!

I would really prefer to have one place to edit right near the functions, and tutorials for full workflows. Can we at least try that and only shift to this approach if it's obviously bad in some way?

asinghvi17:
If the problem is having a place to edit near the function - should we switch to GeometryOps' literate style for Rasters as well?  It would involve a bit of moving code around, giving each method its own file, etc. - but eventually you get all the benefits of long explanations and tutorials, and everything is also centralized.

rafaqz:
I'm happy for literate docs in GeometryOps.jl! I will learn them eventually. But I find the mental overhead of a translation layer a bit much. For example I haven't fixed basic Literate.jl docs issues at Interfaces.jl because I don't immediately understand the problem.

But my problem here is specifically having two places for all the main methods here rather than just one. I really want it to be just one Docs thing per method. Both for maintenance, and I also worry that will be confusing for users.

So we should stretch the current simple API docs to bursting point before having two places to look at the same thing in different ways. 

I know it's kind of boring, but my job here is to push for improvements to the boring old things we already have, instead of new things :)

asinghvi17:
It'd be very hard to understand a reference table of resample methods without pictures.  So I'd argue that in this specific case, we do need a longer tutorial.

asinghvi17:
But maybe this can go in geocompjl, and we can link to it...

rafaqz:
Can't you have pictures tho? The other methods do

lazarusA:
incorporated in https://github.com/rafaqz/Rasters.jl/pull/844

PR #802, opened 2024-10-14
Add get_scale definition to GRIBDatasetsExt

asinghvi17:
Same as for ZarrDatasetsExt.  Needs a test with a grib file.

### DIFF
--- a/ext/RastersGRIBDatasetsExt/gribdatasets_source.jl
+++ b/ext/RastersGRIBDatasetsExt/gribdatasets_source.jl
@@ -11,6 +11,12 @@ function RA._open(f, ::GRIBsource, filename::AbstractString; write=false, kw...)
     RA._open(f, GRIBsource(), ds; kw...)
 end
 
+@inline function RA.get_scale(metadata::RA.Metadata{<: GRIBsource}, scaled::Bool)
+    scale = scaled ? get(metadata, "scale_factor", nothing) : nothing
+    offset = scaled ? get(metadata, "add_offset", nothing) : nothing
+    return scale, offset
+end
+
 RA._sourcetrait(::GDS.Variable) = GRIBsource()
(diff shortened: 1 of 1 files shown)


asinghvi17:
all the segfaults from GDAL, lol...

rafaqz:
Aah it's GDAL warp segfaulting I was hoping it would just go away 

 @visr do you have an idea what's going on there? Has anything obvious changed in the GDAL side?

visr:
Hmm strange. The GDAL build was broken last week: https://github.com/JuliaGeo/GDAL.jl/issues/179
So two new builds were released: https://github.com/JuliaBinaryWrappers/GDAL_jll.jl/releases
It looks like you are on the latest release. Can you easily try [301.901.201](https://github.com/JuliaBinaryWrappers/GDAL_jll.jl/releases/tag/GDAL-v301.901.201%2B0)? Then we could narrow it down. The GDAL 3.9.3 release that came out today also lists two gdalwarp improvements, though this looks more like a regression: https://github.com/OSGeo/gdal/blob/v3.9.3/NEWS.md

We should also fix the Rasters.jl GDAL.jl integration tests again. I haven't looked at them much since they've been red since the beginning.

visr:
So it seems from #806 that GDAL_jll v301.901.201 works ok but v301.902.200 segfaults on gdalwarp, but otherwise seems functional?

rafaqz:
Yeah, all the other GDAL tests are passing.

visr:
Might be a GDAL issue then, not a build failure. Though you could still try the latest GDAL patch release that came out today in the GDAL_jll. Not sure which of the gdalwarp tests is segfaulting. GDAL.jl does have a basic gdalwarp test as well, not as many variations as Rasters.jl though.

rafaqz:
Was there a test for this?

PR #803, opened 2024-10-14
Make extracting `grid_mapping` from CDM datasets safer

asinghvi17:
From conversation with Felix, it turns out that grid mapping is supposed to be stored in global metadata under a single name.  But in the worst case, we should also pick it up from the scattered top-level attributes if people splat it.  This needs a CFCRS.jl package to take the cake...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ defaults:
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    permissions:
+      actions: write
+      contents: read
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
@@ -26,6 +29,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
+      - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
         env:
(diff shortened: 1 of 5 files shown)


rafaqz:
Nice! Maybe we can add a CF firmat to GeoformatTypes for now so we can wrap it in that, then work out how to actually read it later.

asinghvi17:
Yeah I think a `CFCRS{T} where T <: AbstractDict` should work for this...

asinghvi17:
https://github.com/rafaqz/Rasters.jl/blob/netcdf_crs/src/sources/ncdatasets.jl#L407-L508 seems like a good starting point to me, we'll probably need to implement some amount of correction though when translating to PROJJSON.  I might try to spin up a CFCRS.jl this afternoon on that basis.

asinghvi17:
Translating to Proj-strings is probably easiest, then we can work on PROJJSON.

asinghvi17:
I've just tested this on my usecases - it definitely lets Rasters not error out if it doesn't encounter a grid mapping variable.  Saving CFCRS for tomorrow I think this PR should be mergeable after a test.

Will add a bogus NetCDF file with a bad grid mapping structure.

rafaqz:
I think we need to put the resample test last in runtests for now so we can see this pass

rafaqz:
Oops I was not aware Zarr wasn't running!

rafaqz (approved):

asinghvi17:
Holy crap, this actually picked up the CRS!

```julia
raster ┤
  extent: Extent(X = (0.125f0, 359.875f0), Y = (-89.875f0, 89.875f0), zlev = (0.0f0, 0.0f0), Ti = (Dates.DateTime("1981-09-01T12:00:00"), Dates.DateTime("2022-11-08T12:00:00")))
  missingval: missing
  crs: EPSG:4326
  mappedcrs: EPSG:4326
  filename: 
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

```

I have no clue what or how it did it lol, but apparently it works?!

PR #804, opened 2024-10-14
export GFT.CRS too

asinghvi17:
so people can construct WKT more easily


rafaqz:
Where do they need it exactly? I'm hesitant to export that as Proj also has it

asinghvi17:
Ah I didn't think of that.

Does ESRIWellKnownText work equivalently?

rafaqz:
I think it should.

In practice none of them really need CRS explicitly because Mixed mostly works fine.

PR #807, opened 2024-10-15
Refactor tests - use global testset + run Aqua / extension tests last

asinghvi17:
This means that tests will keep running even if one fails, and running Aqua last means you can at least inspect the results of tests.


rafaqz:
I kinda like Aqua first... closing as there are two many stale PRs, reopen if youre really keen

PR #810, opened 2024-10-23
bugfix to _without_mapped_crs

tiemvanderdeure:
just a small bugfix - mappedcrs is an actual `GeoFormat` in this case and we shouldn't try to call anything with it - or you will get this very confusing error (in my case, when calling `crop`)

```
ERROR: MethodError: objects of type EPSG{1} are not callable
The object of type `EPSG{1}` exists, but no method is defined for this combination of argument types when trying to treat it as a callable object.
```

### DIFF
--- a/test/methods.jl
+++ b/test/methods.jl
@@ -556,11 +556,16 @@ end
          missing 1.0     missing]
 
     r_fwd = Raster(A, (X(1.0:1.0:3.0), Y(1.0:1.0:3.0)); missingval=missing)
+    r_crs = Raster(A, (X(1.0:1.0:3.0), Y(1.0:1.0:3.0)); missingval=missing, crs = EPSG(4326))
+    r_mcrs = Raster(A, (X(1.0:1.0:3.0), Y(1.0:1.0:3.0)); missingval=missing, crs = EPSG(4326), mappedcrs = EPSG(4326))
     r_revX = reverse(r_fwd; dims=X)
     r_revY = reverse(r_fwd; dims=Y)
+    st1 = RasterStack((a = r_fwd, b = r_fwd))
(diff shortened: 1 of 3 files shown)


rafaqz:
We should probably try to hit this path in a test ;)

tiemvanderdeure:
Why did I think I was going to get away with not adding tests.

Turns out crop/extend/trim wasn't tested on RasterStacks at all, so I added a bunch more. 

I added some utility functions to be able to do this within the loop - I hope they aren't all too opaque.

tiemvanderdeure:
Methods tests pass - it's just segfaulting again

test/methods.jl
```diff
@@ -574,55 +579,55 @@ end
...
+        @test map(identicalelements, maybe_layers.((extended, extended1, replace_missing(extended_d), ga))...) |> all
+        @test map(identicalelements, maybe_layers.((extended_r, ga_r))...) |> all
         @test all(map(==, lookup(extended_d), lookup(extended)))
+    end
 
-        @testset "unformatted dimension works in crop" begin
-            xdim, ydim = X(1.0:0.2:2.0), Y(1.0:1:2.0)
-            A = Raster(rand((X(1.0:0.2:4.0), ydim)))
-            @test lookup(crop(A; to=xdim), X) == 1.0:0.2:2.0
+    @testset "unformatted dimension works in crop" begin
```
rafaqz, line 593: Indenting changes mean I can't see what you changed

rafaqz (approved):

test/methods.jl
tiemvanderdeure: It's because all of these testsets were inside a for loop unnecessarily, so I took them out of the for loop.

tiemvanderdeure: I didn't change anything about the code other than taking it out of the the loop

Issue #812, opened 2024-11-04
Wrong results from `sum(skipmissing(raster))`


tiemvanderdeure:
I was getting some mysterious results from summing over a raster of gridded population counts from WorldPop (which can be downloaded [here](https://hub.worldpop.org/geodata/summary?id=24777))

This is a big raster, with Float32 data and a Float32 missingval. The output of `sum(skipmissing(x))` is not identical to the sum after aggregating, or the sum of the sum of the slices, or the sum of `collect(skipmissing(x))`

In the worldpop data, the result is off by like 5 percent, but when I try to reproduce I can only get it off by like 0.1%. 
MWE:
```
n = 1000000
mval = -Inf32
ras = Raster(fill(mval, X(1:10000), Y(1:10000)); missingval = mval)
vals = exp.(rand(Float32, n) .* Float32(10))
ras[1:length(vals)] .= vals
sum(skipmissing(ras)) ≈ sum(vals) # false!
sum(skipmissing(replace_missing(ras))) ≈
sum(sum.(skipmissing.(eachslice(ras; dims = X)))) ≈
sum(vals) # true
```
What is going on here?

lazarusA:
just to add to this a little bit more, I  also noticed that doing

`sum(resample(ras,...., method="sum"))` and `sum(ras)` will give slightly different results.

rafaqz:
Think of the sheer number of floating point operations happening in summing your 10_000 * 10_000 array. Theres a lot of room for error to creep in 

There is a correctness optimisation to reduce the error in that process:
https://en.wikipedia.org/wiki/Kahan_summation_algorithm

Maybe we don't hit that path with the Rasters inbuilt `SkipMissingVal` iterator. Or something like that.

There could be specialised methods of `sum` that we need to add? If someone could track down what the difference is, we can add them.

rafaqz:
But seems we don't actually use kahan summation, but pairwise. And not on a a regular iteration:

https://discourse.julialang.org/t/kahan-summation-in-sum/102723

tiemvanderdeure:
Yes it only affects really big rasters, an error of more than 5% seemed crazy, but I also haven't been able to reproduce it.

asinghvi17:
Could you try reproducing this on the original data using AccurateArithmetic.jl?

rafaqz:
For me it was a fraction of a percent, but still big. But its a real problem, this all the algorithms to fix it.

[...]

PR #813, opened 2024-11-05
implement mapreduce for `SkipMissingVal`

tiemvanderdeure:
The implementations is copy-pasted from Base, except that `ismissing(ai)` is replaced by `_missing(ai, itr)`

Solves https://github.com/rafaqz/Rasters.jl/issues/812

I'm not sure what this means for lazy rasters and chunked data from diskarrays. We might want to overload `IndexStyle` for those? Let's see if tests pass.

### DIFF
--- a/src/skipmissing.jl
+++ b/src/skipmissing.jl
@@ -53,3 +53,97 @@ function Base.show(io::IO, s::SkipMissingVal)
     show(io, s.x)
     print(io, ')')
 end
+
+
+## This is adapted from https://github.com/JuliaLang/julia/blob/50713ee4a82eb1b5613647cd74b027315f665080/base/missing.jl#L273-L362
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
ArchGDAL still segfaults on some resample.

There is a PR that will fix this in Julia base and make this unnecessary: https://github.com/JuliaLang/julia/pull/52397. So we can choose to wait it out.

rafaqz:
Oh maybe we can wait for base iterators to be fixed? This will break DiskArrays. 

And yes, it still segfaults.

rafaqz:
Lets close and wait for Base

asinghvi17:
Should we recommend some more accurate summation algorithms in the meantime?  Maybe in the `zonal` docs?

rafaqz:
Yeah possibly. Maybe when we allow online stats we can recommend that?

Issue #816, opened 2024-11-07
`mask` fails when geometry encompasses raster


asinghvi17:
MWE:
```julia
using Rasters, GeoInterface
geom = GeoInterface.Polygon([GeoInterface.LinearRing([(0.0, 0.0), (0.0, 10.0), (10.0, 10.0), (10.0, 0.0), (0.0, 0.0)])])
raster = Raster(ones(11, 11), (X(1:0.1:2), Y(1:0.1:2)))

mask(raster, with=geom)
```
```julia
╭─────────────────────────────────────────╮
│ 11×11 Raster{Union{Missing, Float64},2} │
├─────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Float64} 1.0:0.1:2.0 ForwardOrdered Regular Points,
  → Y Sampled{Float64} 1.0:0.1:2.0 ForwardOrdered Regular Points
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── raster ┤
  extent: Extent(X = (1.0, 2.0), Y = (1.0, 2.0))
  missingval: missing
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 ↓ →  1.0       1.1       1.2       1.3       1.4       1.5       1.6       1.7       1.8       1.9       2.0
 1.0   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.1   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.2   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.3   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.4   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.5   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.6   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.7   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.8   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 1.9   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
 2.0   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing   missing
```

[...]

PR #817, opened 2024-11-07
Test `zonal` on geometries with extents that go beyond the raster

asinghvi17:
We could probably degrade these tests to `rasterize`.  But IMO it's pretty important to test these edge cases.  

If these tests pass, we can enable tiled zonal on pretty much any chunk size, so you could calculate the max elevation on Greenland using a 30-meter cell size DEM like Copernicus, all on your local laptop.

This spawned #816.

### DIFF
--- a/src/methods/burning/edges.jl
+++ b/src/methods/burning/edges.jl
@@ -42,6 +42,7 @@ _x_at_y(e::Edge, y) = (y - e.start[2]) * e.gradient + e.start[1]
 struct Edges <: AbstractVector{Edge}
     edges::Vector{Edge}
     max_ylen::Int
+    min_y::Int
     edge_count::Int
 end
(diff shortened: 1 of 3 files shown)


rafaqz:
Yeah the edge cases are important, but `zonal` is very abstract and hardly does anything. 

Its not actually `rasterize` that matters either, `rasterize` is a really complicated set of tools for writing a wide range of things into rasters where they intersect geometries in some way or other. 

`zonal` is just `mask`, which is `boolmask`, which internally is `burn_geometry!`

Probably testing `boolmask!` more is the cleanest thing. But we could aslo just test `burn_geometry!` directly.

rafaqz:
Was an edge search optimization breaking things, guess I slightly misunderstood the space when I wrote it or just made a dumb mistake.

I think this is solved with very little additional overhead, we just need to look for off-grid points early in the scan or we lose the start line and don't know we are inside a polygon.

Hopefully there is nothing this bad left around.

PR #820, opened 2024-11-12
[WIP] Add a `bylayer` keyword to `zonal`

asinghvi17:
Successor to #775.  This provides a switch that allows you to iterate directly over the values of a rasterstack, instead of iterating over the layers and then the layer values.

It currently defaults to `false` (iteration over values, which is the new behaviour) but we can switch that to `true` (which is less breaking / the old behaviour).

Needs tests and decisions to be made.  This also fails when skipmissing=true, because it's not implemented for RasterStacks.  From discussion in #775 we still need to decide whether skipmissing on a rasterstack should skip a cell if it has any missing values, or skip a cell if all its values are missing.  Personally, I'm inclined to option 1 :).


rafaqz:
Yeah, lets not break things straight away. The main reason to have it as `true` is so `sum` and `mean` just work as-is.

I think we probably need to move away from returning `NamedTuple` at some point to a similar object that math works on. At that point most workflows wont break when we flip the switch, but until then `bylayer=true` needs to be the default.

Project.toml
```diff
@@ -58,7 +58,7 @@ CommonDataModel = "0.2.3, 0.3"
 ConstructionBase = "1"
 CoordinateTransformations = "0.6.2"
 DataFrames = "1"
-DimensionalData = "0.28.2"
+DimensionalData = "0.29"
```
rafaqz, line 61: This may need other changes, lets do that elsewhere and when we have to break things

src/methods/zonal.jl
```diff
@@ -81,11 +87,17 @@ _zonal(f, x::RasterStackOrArray, of::DimTuple; kw...) =
 # We don't need to `mask` with an extent, it's square so `crop` will do enough.
 _zonal(f, x::Raster, of::Extents.Extent; skipmissing=true) =
     _maybe_skipmissing_call(f, crop(x; to=of, touches=true), skipmissing)
-function _zonal(f, x::RasterStack, ext::Extents.Extent; skipmissing=true)
+function _zonal(f, x::RasterStack, ext::Extents.Extent; skipmissing=true, bylayer = false)
```
rafaqz, line 90:
```suggestion
function _zonal(f, x::RasterStack, ext::Extents.Extent; skipmissing=true, bylayer=false)
```

We follow bluestyle in this establishment ;)

src/methods/zonal.jl
```diff
@@ -104,14 +116,19 @@ function _zonal(f, x::AbstractRaster, ::GI.AbstractGeometryTrait, geom;
     return _maybe_skipmissing_call(f, masked, skipmissing)
 end
 function _zonal(f, st::AbstractRasterStack, ::GI.AbstractGeometryTrait, geom; 
-    skipmissing=true, kw...
+    skipmissing=true, bylayer = false, kw...
```
rafaqz, line 119:
```suggestion
    skipmissing=true, bylayer=false, kw...
```

src/methods/zonal.jl
```diff
@@ -131,6 +148,13 @@ end
 function _alloc_zonal(f, x, geoms, n; kw...)
     # Find first non-missing entry and count number of missing entries
     n_missing::Int = 0
+    # TODO: wrap this in try-catch, so that we can throw an intelligent error.
+    # There are a couple of scenarios where this can fail:
+    # 1. The function cannot accept an empty iterator
```
rafaqz, line 153: Ugh I guess we need an `init` keyword to handle this like `reduce`

src/methods/zonal.jl
```diff
@@ -131,6 +148,13 @@ end
...
+    # 2. If passed a RasterStack, the function may not be able to handle layers.  
+    #    In this case you'll get a method error, where one of the arguments is a NamedTuple
+    #    that is the same as eltype(rs).  We should check for that and throw an informative error,
```
rafaqz, line 156: Would be very slick

src/methods/zonal.jl
asinghvi17: I would rather suggest to users:
```julia
zonal(x -> func(x; init = ...), ...)
```
since propagating a kwarg down is always weird...

src/methods/zonal.jl
asinghvi17: Yeah, will try to see if we can do it!  This is probably not very performance sensitive so try catch overhead should be pretty low here.

src/methods/zonal.jl
rafaqz: Ah right of course. We need `FixKW` in base so  this is easier

asinghvi17:
It's pretty easy to miss in the giant error stack though.

asinghvi17:
Was there any objection to this?  Or should I fix it up?

rafaqz:
No it's great we just forgot about it

PR #821, opened 2024-11-13
Breaking: Ambiguities and DD 0.29

rafaqz:
Fix ambiguities. We need to fix DiskArrays before we can test completely against Base, but this is a good start.

Also updates DD by necessity, we need the latest version to fix `selectindices` ambiguities


src/array.jl
```diff
@@ -105,20 +105,17 @@ function DD.modify(f, A::AbstractRaster)
...
 # DiskArrays methods
 
 DiskArrays.eachchunk(A::AbstractRaster) = DiskArrays.eachchunk(parent(A))
 DiskArrays.haschunks(A::AbstractRaster) = DiskArrays.haschunks(parent(A))
-function DA.readblock!(A::AbstractRaster, dst, r::AbstractUnitRange...)
+DA.readblock!(A::AbstractRaster, dst, r::AbstractUnitRange...) =
     DA.readblock!(parent(A), dst, r...)
-end
-function DA.writeblock!(A::AbstractRaster, src, r::AbstractUnitRange...)
+DA.writeblock!(A::AbstractRaster, src, r::AbstractUnitRange...) =
```
asinghvi17, line 117: Can this and friends go into the DD DiskArrays extension, now that it exists?

rafaqz: I'm not sure readblock or writeblock even get hit. But we can delete `eachchunk` and `haschunks`

PR #825, opened 2024-12-01
Maybe breaking? use StepRangeLen instead of LinRange

rafaqz:
Starting this again, hopefully it just passes


asinghvi17:
Could this potentially be the solution to those extent problems I was running into?  Where the extent of a cropped raster was not float64 accurate....

rafaqz:
Possibly, give it a try

PR #826, opened 2024-12-02
construct RasterSeries from AbstractBasicDimArray

tiemvanderdeure:
Just a small quality of life improvement that makes it possible to get a `RasterSeries` from a `DimArray` using a pipe.

Things works nicely together with `@d` to make things like this possible

```julia
using Rasters, RasterDataSources
function get_climate_data(gcm, ssp)
    Raster(rand(X(1:10)))
end
ssps = [SSP370] |> Dim{:ssp}
gcms = [GFDL_ESM4, MRI_ESM2_0] |> Dim{:gcm}
(@d get_climate_data.(gcms, ssps)) |> RasterSeries |> Rasters.combine
```
which returns a `Raster` with dims `X`, `ssp`, and `gcm`

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -52,6 +52,8 @@ DD.modify(f, A::AbstractRasterSeries) = map(child -> modify(f, child), values(A)
     RasterSeries(paths::AbstractArray{<:AbstractString}, dims; child, duplicate_first, kw...)
     RasterSeries(path:::AbstractString, dims; ext, separator, child, duplicate_first, kw...)
 
+    RasterSeries(objects::AbstractBasicDimArray; kw...)
+
 Concrete implementation of [`AbstractRasterSeries`](@ref).
 
 A `RasterSeries` is an array of `Raster`s or `RasterStack`s, along some dimension(s).
--- a/src/series.jl
+++ b/src/series.jl
@@ -116,6 +116,8 @@ struct RasterSeries{T,N,D,R,A<:AbstractArray{T,N}} <: AbstractRasterSeries{T,N,D
     dims::D
     refdims::R
 end
+RasterSeries(data::DD.AbstractBasicDimArray; kw...) = 
+    RasterSeries(data, dims(data); kw...)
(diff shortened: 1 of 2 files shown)


rafaqz:
So slick.

Especially with RDS climate types as lookups.

src/series.jl
```diff
@@ -116,6 +116,8 @@ struct RasterSeries{T,N,D,R,A<:AbstractArray{T,N}} <: AbstractRasterSeries{T,N,D
     dims::D
     refdims::R
 end
+RasterSeries(data::DD.AbstractBasicDimArray; kw...) = 
+    RasterSeries(data, dims(data); kw...)
```
rafaqz, line 120:
```suggestion
RasterSeries(data::DD.AbstractBasicDimArray{<:Union{AbstractRasterStack,AbstractRaster}}; kw...) = RasterSeries(data, dims(data); 
```

Has to hold a Raster or RasterStack

tiemvanderdeure: I didn't want to limit it to that, in principle it could hold strings (of filenames) as well. If it holds something that isn't allowed then `RasterSeries(data, dims(data)` will fail anyway

rafaqz: Ahh right because strings will work too. Ok makes sense

rafaqz (approved):

rafaqz:
Did you want to add docs in this PR? Or just merge?

tiemvanderdeure:
Just merge! I'll make a docs example in a separate PR.

asinghvi17:
Just to clarify here - is a RasterSeries essentially a lazy concatenation of multiple dimarrays of dimension n, along some dimension n+1 (and maybe also n+2 etc)?

rafaqz:
It's basically just a Raster of Rasters or RastersStacks with some helper methods.

It's not a lazy concatenation, but using `Rasters.combine` will concatenate it to a single Raster/RasterStack, and for DiskArrays that will be lazy

PR #831, opened 2024-12-06
Update to SDM workflow example in the docs

tiemvanderdeure:
Uses SDM.jl/MLJ to do the whole workflow.

Also updated a few steps that are a lot less clunky now.

(And replaced plots.jl with makie, but feel free to reverse/improve).


rafaqz:
Thanks, looks good

tiemvanderdeure:
Hm I just took a look at the docs and the output of the last cells isn't shown - and it has `@example` in the top right. I must have made some typo?
 billede

rafaqz:
Huh, I can't see anything obvious

tiemvanderdeure:
Same, and it doesn't error either. It could be something with the SDM package as it's unregistered.

rafaqz:
That shouldn't make a difference. I would guess a missing newline or something.

But also that package really wants registering ;)

Issue #814, opened 2024-11-05
Method ambiguity - indexing with DimSelectors doesn't work on Projected axes


asinghvi17:
MWE:
```julia
r1 = Raster(WorldClim{Climate}, :tmin, month = 1)
r2 = r1[1:10, 1:10]
r3 = r1[DimSelectors(r2, selectors = At)]
```

Error (ignore the task stuff):
```julia

    nested task error: MethodError: selectindices(::Projected{…}, ::At{…}) is ambiguous.
    
    Candidates:
      selectindices(l::Projected, sel::At)
        @ Rasters ~/.julia/packages/Rasters/0qvMe/src/lookup.jl:96
      selectindices(l::DimensionalData.Dimensions.Lookups.Lookup, sel::At{<:AbstractVector}; kw...)
        @ DimensionalData.Dimensions.Lookups ~/.julia/packages/DimensionalData/hhCBb/src/Lookups/selector.jl:140
    
    Possible fix, define
      selectindices(::Projected, ::At{<:AbstractVector})
    
    Stacktrace:
      [1] _dims2indices
        @ ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:117 [inlined]
      [2] _dims2indices
        @ ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:109 [inlined]
      [3] macro expansion
        @ ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:55 [inlined]
      [4] _dims2indices(lookups::Tuple{Projected{…}, Projected{…}}, dims::Tuple{X{…}, Y{…}}, I::Tuple{X{…}, Y{…}})
        @ DimensionalData.Dimensions ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:55
      [5] dims2indices
        @ ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:50 [inlined]
      [6] dims2indices
        @ ~/.julia/packages/DimensionalData/hhCBb/src/Dimensions/indexing.jl:32 [inlined]
      [7] _dim_dotview
        @ ~/.julia/packages/DimensionalData/hhCBb/src/array/indexing.jl:110 [inlined]
      [8] dotview(::Raster{…}, ::Y{…}, ::X{…})
        @ DimensionalData ~/.julia/packages/DimensionalData/hhCBb/src/array/indexing.jl:87
[...]

PR #832, opened 2024-12-09
`open` all datasets at `zonal` entry point

asinghvi17:
this should massively increase performance for `lazy=true` rasters.  But it needs tests with rasters of every form.


rafaqz:
Yeah probably a lazy test is needed, there may be one already, it can go with GDAL zonal tests

rafaqz:
We really need chunked burning for this to have much of an effect.

asinghvi17:
> We really need chunked burning for this to have much of an effect.

Doesn't that already work with the exception of https://github.com/rafaqz/Rasters.jl/issues/816?  Then the chunking only has to be done on the top level...

rafaqz:
I'm pretty sure #814 is fixed now. 

But there will be other memory problems from not chunking

PR #833, opened 2024-12-10
open all rasters in `extract` too

asinghvi17:
same thing as the last PR.  Since this is inside Rasters it's probably OK to use `open` this way?  Assuming that all output arrays are pre-allocated (not views).


rafaqz:
Thanks.

I'm afraid extract needs a lot more work to be good on lazy data. Probably using your RangeExtractor or something to read the required chunks before extracting.

PR #835, opened 2024-12-10
Better Raster and RasterStack constructors

rafaqz:
This PR does two main things:
- allow `Raster(extent; res)` like terra etc
- allow `Raster(dataset)` / `Raster(variable)` / `RasterStack(dataset)` for all backends, so that ArchGDAL and CommonDataModel objects can be used more easily as part of a Rasters.jl workflow.


asinghvi17:
Awesome, this is exactly what I needed

rafaqz:
Which part?

asinghvi17:
Raster(dataset) / RasterStack(dataset)

rafaqz:
Yeah its actually pretty weird that didn't work earlier. Im also wondering if we need the reverse - like `CommonDataModel.variable(raster, [name])` and `CommonDataModel.dataset(raster)` should probably work.

asinghvi17:
As long as we only want that working on CDM like datasets, that makes sense.

It occurs that this is probably also the way to do `lazy = :open` - another interactive opening API, and maybe a mutable struct wrapper that can wrap a finalizer

rafaqz:
Yeah, a bunch of the work here will help with `:open` too. Don't worry that is still the plan! will just take a while longer still.

I'm less worried about the finalizer as it is mostly handled by other packages, and more about the ability to manually `close` files when you need to, like when you hit the filesystem file limit.

asinghvi17:
I guess we can assume that if the user pulls out the dataset before we call `close`, it's their problem?

But in general I don't think e.g. ZarrDatasets.jl has a destructor.  Of course ZarrDatasets doesn't maintain file pointers so it's pointless to destruct there, but I guess NCDatasets will maintain a pointer to the (open) file via libnetcdf.  And it seems to have a `close(ds)` function so that works.

I guess this will work like `open` and `close` do in Base, exactly the same interface actually...so we only need to define `RA._open` and `RA._close` on each backend object?  Or something similar.

PR #839, opened 2024-12-24
Faster `mosaic`

rafaqz:
Known reducing functions can be written as whole regions at a time, as in `rasterize`.

Generic functions can also be optimized a lot from the current very old and type unstable implementation


src/methods/mosaic.jl
```diff
@@ -145,18 +145,24 @@ Combine `regions` in `Raster` or `RasterStack` `x` using the function `f`.
...
+Cut out scandinavian countries and plot:
 
 ```jldoctest
-using Rasters, RasterDataSources, ArchGDAL, Statistics, Plots
-st = read(RasterStack(WorldClim{Climate}; month=1))
-aus = st[X=100.0 .. 160.0, Y=-50.0 .. -10.0]
-africa = st[X=-20.0 .. 60.0, Y=-40.0 .. 35.0]
-mosaic!(first, st, aus, africa)
-plot(st)
+using Rasters, RasterDataSources, NaturalEarth, DataFrames
```
lazarusA, line 151: I think you also need `using Dates`, since you are asking for `July`

PR #840, opened 2024-12-27
fix features links

lazarusA:
small fixes and docs cleanup.


lazarusA:
I'm working on `resample` (using it again for my own work 😄 ) and probably it should be best to use this opportunity to combine and finish #730 + #801 (as a tutorial) and #809  . Not in this PR, but in the next one.

rafaqz:
Yeah there are a few half done PRs building up it would be nice to incorporate them

PR #844, opened 2024-12-28
resample cellarea continue

lazarusA:
getting this #800 #801 and #730  in.

This is getting too long. So, I will merge the current changes, and then we can refine later if needed.  Note the following:

- [x] In the `Manual` section I added all methods and data sources. Currently, the only method  pointing to a new section is `cellarea` since is the only one that has additional documentation. The remaining ones, point out to the API docs. Hence, if we need more docs on the different methods, please let's first add the API docs at the top of each section, before going into more details. 
- [x] links to methods (there are not that nice), hence populating with extended docs would be good (we could shorten them then- being also maybe more stable in the long run).
- [x] this PR contains a tutorial on `resample` and reprojection with `resample` and ArchGDAL internals for the transformations. 
- [x] another reorganization of topics, trying to be consistent with things that should go into the manual (base functions) and tutorials (a combinations of them to create workflows)


docs/src/manual/cellarea.md
```diff
@@ -0,0 +1,71 @@
...
+We can now construct the x and y lookups.  Here we'll use a start-at-one, step-by-five grid.
+Note that we're specifying that the "sampling", i.e., what the coordinates actually mean, 
+is `Intervals(Start())`, meaning that the coordinates are the starting point of each interval.
+
+This is in contrast to `Points()` sampling, where each index in the raster represents the value at a sampling point;
+here, each index represents a grid cell, which is defined by the coordinate being at the start.
+
+````@example cella
+x = X(1:5:30; sampling = Intervals(Start()), crs = EPSG(4326))
+y = Y(50:5:80; sampling = Intervals(Start()), crs = EPSG(4326))
```
asinghvi17, line 43: This doesn't work without `using Rasters.Lookups`.

docs/src/manual/cellarea.md
```diff
@@ -0,0 +1,71 @@
...
+
+````@example cella
+using Makie, CairoMakie# , GeoMakie
+heatmap(ras; axis = (; aspect = DataAspect()))
+````
+
+We can also try this using the planar method, which simply computes the area of the rectangle using `area = x_side_length * y_side_length`:
+
+````@example cella
+cellarea(ras, Planar())
```
asinghvi17, line 68:
```suggestion
cellarea(Planar(), ras)
```

docs/src/manual/resample_warp.md
```diff
@@ -0,0 +1,67 @@
...
+**[`resample`](@ref)** "re-samples" the 
+data by interpolation and can also aggregate or disaggregate, changing the resolution.
+It always returns a `Regular` lookup (like a range), and is the most flexible of the 
+resampling methods.
+
+This uses GDAL's `gdalwarp` algorithm under the hood.  You can call that via [`warp`](@ref)
+if you need more control, but generally `resample` is sufficient. 
+
+Rasters.jl has a few other methods to change the lookups of a raster.  These are:
+- [`reproject`](@ref), which simply directly reprojects the lookup axes
```
asinghvi17, line 25:
```suggestion
- [`reproject`](@ref), which directly reprojects the lookup axes

File: docs/src/manual/resample_warp.md [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+
+Rasters.jl has a few other methods to change the lookups of a raster.  These are:
+- [`reproject`](@ref), which simply directly reprojects the lookup axes 
+  (but is **only usable for specific cases**, where the source and destination 
+  coordinate systems are both cylindrical, like the long-lat, Mercator, or Web-Mercator projections.) 
+
+  This is a lossless operation and keeps the data exactly the same - only the axes are changed. 
+
+- [`aggregate`](@ref) and [`disaggregate`](@ref), which change the resolution of 
+  the raster by clumping ([`aggregate`](@ref)) or splitting ([`disaggregate`](@ref)) cells.
```
asinghvi17, line 32:
```suggestion
  the raster by merging ([`aggregate`](@ref)) or splitting ([`disaggregate`](@ref)) cells.

File: docs/src/resample_proj.md [hunk 5]
```diff
@@ -0,0 +1,124 @@
...
+````
+
+now, let's go back to `latitude` and `longitude`
+
+````@example modis
+# ? also here, do we need to shift X, Y?
+ras_epsg = resample(ras_sin; size=(1440,720), crs=EPSG(4326), method="sum")
+````
+
+and let's apply `shiftlocus` such that we can harmonize coordinates, which might be needed when building bigger datasets:
```
asinghvi17, line 60: what's "harmonizing coordinates"?  Maybe this should be explained as 
```suggestion
and let's apply `shiftlocus` such that the lookups share the exact same grid, which might be needed when building bigger datasets:
```

docs/src/resample_proj.md
```diff
@@ -0,0 +1,124 @@
...
+
+create the raster
+
+````@example modis
+ras_scratch = Raster(ra_data, (X(x_range; sampling=Intervals(Start())), Y(y_range; sampling=Intervals(Start()))),
+    crs=SINUSOIDAL_CRS)
+````
+
+::: warning
+At the moment, you need to specify `sampling=Intervals(Start())` for `X` and `Y`.
```
asinghvi17, line 93:
```suggestion
At the moment, you need to specify `sampling=Intervals(Start())` for `X` and `Y`.

This requires that you run `using Rasters.Lookups`, where the `Intervals` and `Start` types are defined.

File: docs/src/resample_proj.md [hunk 7]
```diff
@@ -0,0 +1,124 @@
...
+````
+
+and compare the total counts again!
+
+````@example modis
+nansum(ras_m), nansum(locus_resampled)
+````
+
+::: danger
+Note that all counts are a little bit off. Could we mitigate this a some more?
```
asinghvi17, line 120:
```suggestion
Note that all counts are a little bit off. Could we mitigate this some more?

asinghvi17 (commented):
Thanks for the port!  Added a few suggestions.

docs/src/manual/cellarea.md
lazarusA: I will still see what goes into the docstring. Trying not to duplicate things, as suggested by @rafaqz .

asinghvi17: How do you mean?  Without that this example doesn't work...since `Intervals` and `Start` are not loaded into the main namespace

Issue #851, opened 2025-01-07
Rasterize with ext and res uses wrong corners of the Extent


felixcremer:
When I am running rasterize with a extent as `to` and a resolution I would have expected that the X axis spans the range of the X extent and the Y axis range of extent.Y. It seems that the X-axis is constructed from the first of the entries in extent and Y from the last entries. 
In the example below I would have expected that X has values `-1:0.01:1` and Y `0.33:0.01:2`

```julia
julia> extr =Extent(X=(-1,1.1), Y=(0.33,2));

julia> rasr = rasterize(prand, to=extr, res=0.1, fill=1, reducer=sum)
┌ 14×10 Raster{Union{Missing, Int64}, 2} nothing ┐
├────────────────────────────────────────────────┴─────────────────────────────────────────── dims ┐
  ↓ X Projected{Float64} -1.0:0.1:0.3 ForwardOrdered Regular Intervals{Start},
  → Y Projected{Float64} 1.1:0.1:2.0 ForwardOrdered Regular Intervals{Start}
├──────────────────────────────────────────────────────────────────────────────────────── metadata ┤
  Metadata of Dict{Any, Any}()
├────────────────────────────────────────────────────────────────────────────────────────── raster ┤
  extent: Extent(X = (-1.0, 0.4), Y = (1.1, 2.1))
  missingval: missing
└──────────────────────────────────────────────────────────────────────────────────────────────────┘
 ⋮      ⋱  
```

asinghvi17:
That is indeed awkward...probably a typo?

tiemvanderdeure:
While https://github.com/rafaqz/Rasters.jl/pull/852 mostly fixes this, maybe there is still room for improvement on how edges are handled.

```
point = (0.5,0.5)
extr =Extent(X=(-1,1.1), Y=(0.0,2.0));
rasr = rasterize(point, to=extr, res=0.1, fill=1, reducer=sum)
extent(rasr)
```

Gives `Extent(X = (-1.0, 1.2000000000000002), Y = (0.0, 2.1))`. So it's like 1 row and 1 column more than I would expect. 

This happens in `_extent2dims`

rafaqz:
We need extents.jl to have closed or open intervals.

tiemvanderdeure:
Fixed now!

```julia-repl
julia> using Rasters, Extents

julia> extr =Extent(X=(-1,1.1), Y=(0.33,2));

[...]

PR #852, opened 2025-01-07
Fix `_extent` when called on an extent

tiemvanderdeure:
Solves https://github.com/rafaqz/Rasters.jl/issues/851

### DIFF
--- a/src/methods/burning/extents.jl
+++ b/src/methods/burning/extents.jl
@@ -50,7 +50,7 @@ function _extent(::GI.AbstractFeatureCollectionTrait, features; kw...)::XYExtent
     end
     return _float64_xy_extent(ext)
 end
-_extent(ext::XYExtent; kw...) = ext
+_extent(ext::XYExtent; kw...) = _float64_xy_extent(ext)
 
 function _float64_xy_extent(ext::Extents.Extent)
     xbounds = map(Float64, ext.X)
--- a/src/methods/burning/extents.jl
+++ b/src/methods/burning/extents.jl
@@ -1,18 +1,20 @@
 const XYExtent = Extents.Extent{(:X,:Y),Tuple{Tuple{Float64,Float64},Tuple{Float64,Float64}}}
 
 # Get the bounds of a geometry
-_extent(geom; kw...)::XYExtent = _extent(GI.trait(geom), geom; kw...)
-function _extent(::Nothing, data; geometrycolumn, kw...)::XYExtent
+function _extent(data; geometrycolumn=nothing, kw...)::XYExtent 
     geoms = _get_geometries(data, geometrycolumn)
-
+    _extent(GI.trait(geoms), geoms)
+end
+function _extent(::Nothing, geoms)::XYExtent
+    # because geoms was returned from _get_geometries, it must be an iterable of valid geometries
     g1 = first(geoms)
     if GI.trait(g1) isa GI.PointTrait 
         xs = extrema(p -> GI.x(p), geoms)
         ys = extrema(p -> GI.y(p), geoms)
         return _float64_xy_extent(Extents.Extent(X=xs, Y=ys))
     else
-        ext = reduce(geoms; init=_extent(first(geoms))) do ext, geom
-            Extents.union(ext, _extent(geom; kw...))
+        ext = reduce(geoms; init=_extent(GI.trait(g1), g1)) do ext, geom
+            Extents.union(ext, _extent(GI.trait(geom), geom))
         end
         return _float64_xy_extent(ext)
     end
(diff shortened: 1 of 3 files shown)


rafaqz:
Want to add a test now?

src/methods/burning/extents.jl
```diff
@@ -50,6 +50,7 @@ function _extent(::GI.AbstractFeatureCollectionTrait, features; kw...)::XYExtent
     end
     return _float64_xy_extent(ext)
 end
+_extent(ext::XYExtent; kw...) = ext
```
rafaqz, line 53:
```suggestion
_extent(ext::XYExtent; kw...) = _float64_xy_extent(ext)
```

src/methods/burning/extents.jl
```diff
@@ -50,6 +50,7 @@ function _extent(::GI.AbstractFeatureCollectionTrait, features; kw...)::XYExtent
     end
     return _float64_xy_extent(ext)
 end
+_extent(ext::XYExtent; kw...) = _float64_xy_extent(ext)
```
rafaqz, line 53:
```suggestion
_extent(ext::Extent; kw...) = _float64_xy_extent(ext)
```

src/methods/burning/extents.jl
```diff
@@ -50,6 +50,7 @@ function _extent(::GI.AbstractFeatureCollectionTrait, features; kw...)::XYExtent
     end
     return _float64_xy_extent(ext)
 end
+_extent(ext::Extent; kw...) = _float64_xy_extent(ext)
```
tiemvanderdeure, line 53: I thought about this one, but do we really want to force the convert to float64? This will mean that `rasterize` with an extent in integers or float32 will return float64 dimensions

test/methods.jl
```diff
@@ -762,3 +762,10 @@ test = rebuild(ga; name = :test)
     @test_throws "strictly positive" Rasters.sample(StableRNG(123), test, 3, skipmissing = true, replace = false)
     @test_throws "Cannot draw" Rasters.sample(StableRNG(123), test, 5, replace = false)
 end
+
+@testset "extent" begin
+    ga = Raster(A, (X(1.0:1:2.0), Y(1.0:1:2.0)); missingval=missing) 
+    ext = extent(ga)
+    @test ext === Extent(X=(1.0,2.0), Y=(1.0,2.0))
+    @test extent(ext) === ext
```
rafaqz, line 770:
```suggestion
    @test Rasters._extent(ext) === ext
```
Shouldn't it be something like this?

tiemvanderdeure: You're right, it should be.

src/methods/burning/extents.jl
rafaqz: We convert everything else already. I think there were 2 reasons - accuracy of ext + res/size is much better with Float64, and it reduces compilation for force the extent to always be the same object.

But we can change everything back to allow `Float32` in another PR if you think its necessary

tiemvanderdeure: I think we want `_extent(ext::Extent; kw...) = Extents.Extent(X=ext.X, Y=ext.Y)`. If `ext` doesn't have an X or Y field it will give a reasonable error message.

tiemvanderdeure: Okay if it gets converted elsewhere then nevermind

rafaqz: We actually want to force `::XYExtent` on the method to be consistent

rafaqz: We can have the Float32 converstion elsewhere but lets conform with the current setup for now

felixcremer: Yeah, but without that my actual example failed when I used a Float32 extent.

test/methods.jl
```diff
@@ -762,3 +762,10 @@ test = rebuild(ga; name = :test)
     @test_throws "strictly positive" Rasters.sample(StableRNG(123), test, 3, skipmissing = true, replace = false)
     @test_throws "Cannot draw" Rasters.sample(StableRNG(123), test, 5, replace = false)
 end
+
+@testset "extent" begin
+    ga = Raster(A, (X(1.0:1:2.0), Y(1.0:1:2.0)); missingval=missing)
```
felixcremer, line 767: We should also test this for an Extent, that is based on float32

src/methods/burning/extents.jl
tiemvanderdeure: Okay I didn't realize `XYExtent` has the Float64 type hard-coded in. Guess that is needed for type stability?

rafaqz: I think type stability was one reason, this recursive calls were not compiling away and forcing the type fixes that

tiemvanderdeure:
Just noticed there were a few places where keywords (specifically `geometrycolumn`) weren't forwarded but should have been so I squeezed that in.

test/methods.jl
rafaqz: @tiemvanderdeure we need this to merge

tiemvanderdeure: I thought you said you wanted to fix that in a separate PR? Right now this would fail, I guess?

rafaqz: Will it? I don't totally understand, won't it just convert to Float64?

We just need to check the input works

tiemvanderdeure:
We just need https://github.com/rafaqz/Rasters.jl/pull/853 to merge to address the test failure.

tiemvanderdeure:
Tests pass, I think this is ready to merge

PR #853, opened 2025-01-08
handle `FeatureTrait` and `GeometryCollectionTrait` in  `_get_geometries`

tiemvanderdeure:
An `ArchGDAL.Feature` doesn't have a `GI.trait`, but returns its geometry with `GI.geometry`. `_get_geometries` should try that before anything else.

I'll wait for https://github.com/rafaqz/Rasters.jl/pull/852 to merge to add tests

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -200,6 +200,10 @@ end
 # get geometries from what may be a table with a geometrycolumn or an interable of geometries
 # if it has no geometry column and does not iterate valid geometries, error informatively
 function _get_geometries(data, ::Nothing)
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
Actually this should probably just be fixed in ArchGDAL. Surely ArchGDAL.Feature should have `GI.FeatureTrait`. Then in _get_geometries we should check for FeatureTrait, like we already do for FeatureCollectionTrait

tiemvanderdeure:
This should fix the test failure in https://github.com/rafaqz/Rasters.jl/pull/852

PR #856, opened 2025-01-09
Allow `reproject(obj, target_crs)` as well

asinghvi17:
Just allows a different argument order so that Rasters and GO reproject definitions are compatible

### DIFF
--- a/src/methods/reproject.jl
+++ b/src/methods/reproject.jl
@@ -22,7 +22,7 @@ $CRS_KEYWORD
 """
 reproject(x; crs::GeoFormat) = reproject(crs, x)
 reproject(x, target::GeoFormat) = reproject(target, x)
-reproject(::GeoFormat, ::GeoFormat) = throw(ArgumentError("You need to provide a raster object to reproject. Got two coordinate reference systems.")
+reproject(::GeoFormat, ::GeoFormat) = throw(ArgumentError("You need to provide a raster object to reproject. Got two coordinate reference systems."))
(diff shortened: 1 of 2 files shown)


src/methods/reproject.jl
```diff
@@ -21,6 +21,7 @@ are silently returned without modification.
 $CRS_KEYWORD
 """
 reproject(x; crs::GeoFormat) = reproject(crs, x)
+reproject(x, target::GeoFormat) = reproject(target, x)
```
felixcremer, line 24:
```suggestion
reproject(x, target::GeoFormat) = reproject(target, x)
reproject(::GeoFormat, ::GeoFormat) = throw(ArgumentError("You need to provide a raster object to reproject. Got two coordinate reference systems.")
```

src/methods/reproject.jl
```diff
@@ -21,6 +21,8 @@ are silently returned without modification.
...
+reproject(::GeoFormat, ::GeoFormat) = throw(ArgumentError("You need to provide a raster object to reproject. Got two coordinate reference systems.")
```
felixcremer, line 25:
```suggestion
reproject(::GeoFormat, ::GeoFormat) = throw(ArgumentError("You need to provide a raster object to reproject. Got two coordinate reference systems."))
```

felixcremer:
Lgtm

rafaqz:
Maybe we should swap the docstring to the GO format at the same time?

PR #860, opened 2025-01-15
add multidimensional spatial means tutorial

tiemvanderdeure:
I've wanted to fit this into the docs somewhere for a while, maybe this is a reasonable way to do this?

Using dimensions like time period, SSP, GCM has become a standard part of my workflow and I think it's just a really neat way to work with dimensions.

It didn't become quite as slick as I'd hoped, mainly because of some inconsistencies in the worldclim data. I'm open to suggestions (either on the example itself or on where to fit it into the docs).

### DIFF
--- a/docs/src/tutorials/spatial_mean.md
+++ b/docs/src/tutorials/spatial_mean.md
@@ -141,3 +141,54 @@ We've also seen how to use the `cellarea` function to compute the area of each c
 
 We've seen that the spatial mean is not the same as the arithmetic mean, and that we need to account for the area of each cell when computing the average.
 
+## Bonus: Computing spatial means across dimensions 
+
+As a next step, we would like to know how precipitation will change in Chile until the end of the 21st century. To do this, we can use climate model outputs. This data can come from multiple climate models (GCMs) and under different socio-economic scenarios (SSPs). We'll use additional dimensions to keep track of these.
+
+First we define a simple function takes an SSP (socioeconomic scenario) and a GCM (climate model) as input, and downloads the appropriate climate data. 
+
+````@example zonal
+using Dates
+getfutureprec(ssp, gcm) = Raster(WorldClim{Future{Climate, CMIP6, gcm, ssp}}, :prec, date = Date(2090))
+````
+
+We will leverage some tools from [DimensionalData](https://github.com/rafaqz/DimensionalData.jl), which is the package that underlies Rasters.jl. Rather than having a seperate Raster for each combination of GCM and SSP, `gcm` and `ssp` will be additional dimensions, and our Raster will be 4-dimensional (X-Y-gcm-ssp). 
+
+To do this, we first define two dimensions that correspond to the SSPs and GCMs we are interested in, then use the `@d` macro from [DimensionalData](https://github.com/rafaqz/DimensionalData.jl) to preserve these dimensions as we get the data, and then combine all Rasters into a single object using `Rasters.combine`
+
+````@example cellarea
+SSPs = Dim{:ssp}([SSP126, SSP370]) # SSP126 is a low-emission scenario, SSP370 is a high-emission scenario
+GCMs = Dim{:gcm}([GFDL_ESM4, IPSL_CM6A_LR]) # These are different general circulation (climate) models
+
+precip_future = (@d getfutureprec.(SSPs, GCMs)) |> RasterSeries |> Rasters.combine
+````
+
+Since the format of WorldClim's datasets for future climate is slightly different from the dataset for the historical period, this actually returned a 5-dimensional raster, with a `Band` dimension that represents months. Here we'll just select the 6th month, matching the selection above. We will also replace the `NaN` missing value by the more standard `missing` using [`replace_missing`](@ref).
+
+````@example cellarea
+precip_future = precip_future[Band = 6]
+precip_future = replace_missing(precip_future)
+````
+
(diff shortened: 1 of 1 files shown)


rafaqz:
Maybe update to using and merging into the `breaking` branch, a lot of this syntax will be simpler, like not needing `replace_missing`.

Otherwise I will have to fix it all in the next week 😅

docs/src/tutorials/spatial_mean.md
```diff
@@ -141,3 +141,54 @@ We've also seen how to use the `cellarea` function to compute the area of each c
 
 We've seen that the spatial mean is not the same as the arithmetic mean, and that we need to account for the area of each cell when computing the average.
 
+## Bonus: Computing spatial means across dimensions 
+
+As a next step, we would like to know how precipitation will change in Chile until the end of the 21st century. To do this, we can use climate model outputs. This data can come from multiple climate models (GCMs) and under different socio-economic scenarios (SSPs). We'll use additional dimensions to keep track of these.
+
+First we define a simple function takes an SSP (socioeconomic scenario) and a GCM (climate model) as input, and downloads the appropriate climate data.
```
felixcremer, line 148:
```suggestion
First we define a simple function which takes an SSP (socioeconomic scenario) and a GCM (climate model) as input, and downloads the appropriate climate data. 
```

docs/src/tutorials/spatial_mean.md
```diff
@@ -141,3 +141,54 @@ We've also seen how to use the `cellarea` function to compute the area of each c
...
+First we define a simple function takes an SSP (socioeconomic scenario) and a GCM (climate model) as input, and downloads the appropriate climate data. 
+
+````@example zonal
+using Dates
+getfutureprec(ssp, gcm) = Raster(WorldClim{Future{Climate, CMIP6, gcm, ssp}}, :prec, date = Date(2090))
+````
+
+We will leverage some tools from [DimensionalData](https://github.com/rafaqz/DimensionalData.jl), which is the package that underlies Rasters.jl. Rather than having a seperate Raster for each combination of GCM and SSP, `gcm` and `ssp` will be additional dimensions, and our Raster will be 4-dimensional (X-Y-gcm-ssp). 
+
+To do this, we first define two dimensions that correspond to the SSPs and GCMs we are interested in, then use the `@d` macro from [DimensionalData](https://github.com/rafaqz/DimensionalData.jl) to preserve these dimensions as we get the data, and then combine all Rasters into a single object using `Rasters.combine`
```
felixcremer, line 157: Maybe this should link to the docstring of `@d` directly?

docs/src/tutorials/spatial_mean.md
```diff
@@ -141,3 +141,54 @@ We've also seen how to use the `cellarea` function to compute the area of each c
...
+SSPs = Dim{:ssp}([SSP126, SSP370]) # SSP126 is a low-emission scenario, SSP370 is a high-emission scenario
+GCMs = Dim{:gcm}([GFDL_ESM4, IPSL_CM6A_LR]) # These are different general circulation (climate) models
+
+precip_future = (@d getfutureprec.(SSPs, GCMs)) |> RasterSeries |> Rasters.combine
+````
+
+Since the format of WorldClim's datasets for future climate is slightly different from the dataset for the historical period, this actually returned a 5-dimensional raster, with a `Band` dimension that represents months. Here we'll just select the 6th month, matching the selection above. We will also replace the `NaN` missing value by the more standard `missing` using [`replace_missing`](@ref).
+
+````@example cellarea
+precip_future = precip_future[Band = 6]
```
felixcremer, line 169: We could keep the Band information and do the next steps for all months at the same time?

felixcremer (commented):
Nice.

docs/src/tutorials/spatial_mean.md
tiemvanderdeure: We could, I just wanted to keep it simple. At the end of the whole thing it just prints a 2x2 dimArray and that's easier to read than a 12x2x2 dimarray

docs/src/tutorials/spatial_mean.md
rafaqz: Yeah we should try it, Rasters/DD doc links came up on slack too it might make things easier

lazarusA: apparently, we don't support that yet in DVpress (never saw the benefit or was even aware of this,  until now). I might take a look later this week.

asinghvi17: I had some code to do that in my DocumenterMarkdown PR, will see about resurrecting it

docs/src/tutorials/spatial_mean.md
felixcremer: That makes sense and I would add this info to the tutorial:
```
We are restricting here to one Band for simplicity but the analysis would also work for all Bands simultaneously.
```

lazarusA:
@rafaqz @tiemvanderdeure let's go ahead with this. We do a general fix once the Breaking release lands.

rafaqz:
Sure unless @tiemvanderdeure already started on it

tiemvanderdeure:
I haven't made the fix yet - let me just make a few changes to the text and we can merge

tiemvanderdeure:
Good to go for me. I'll fix when the next breaking release hits. (I fixed a little text thing in the first half - the precipitation per square meter in Chile is not 8 * 10^13)

PR #880, opened 2025-01-30
fix the `create` docs

tiemvanderdeure:
such elaborate docs! Would be a shame if no one could read them

### DIFF
--- a/src/create.jl
+++ b/src/create.jl
@@ -124,7 +124,7 @@ RasterStack("created.nc")
 └───────────────────────────────────────────────────────────────────────────────────────────┘
 ```
 """
-create
+function create end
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
Doubt the test failures have much to do with this PR ;)

More likely it's because of https://github.com/rafaqz/DimensionalData.jl/pull/915

src/create.jl
```diff
@@ -124,6 +124,7 @@ RasterStack("created.nc")
 └───────────────────────────────────────────────────────────────────────────────────────────┘
 ```
 """
+create
```
rafaqz, line 127:
```suggestion
function create end
```
Sorry I like the explicit style, for newbies it looks weird as-is...

PR #883, opened 2025-01-31
write to zarr

rafaqz:
@felixcremer I noticed `force=true` is going to be a bit more difficult for zarr given the folders... what do you think it should do?

Also wondering if appending data to an existing array is something zarr can do


felixcremer:
In YAX we have an overwrite keyword and if that is true we delete the whole folder before writing a new one.

felixcremer:
Appending data should work. There is an append keyword in YAXArrays savedataset function. I am not entirely sure what it does exactly

PR #885, opened 2025-02-03
mention benchmark source in README

kadyb:
Hi, can we provide the benchmark source in the README?


rafaqz:
Yes for sure! We should probably get it merged too

PR #886, opened 2025-02-05
add an option to disaggregate using a view

tiemvanderdeure:
This adds a `view` keyword option to `disaggregate`, in which case it just returns a view instead of eagerly extracting the values and allocating a new array.

I think it's worth adding as it's just a few lines of code and minimal feature creep, just one keyword, and whenever memory is an issue this is kind of neat.

We could consider a lazy aggregation as well, though it wouldn't be quite as easy.

### DIFF
--- a/src/methods/aggregate.jl
+++ b/src/methods/aggregate.jl
@@ -392,6 +392,8 @@ end
     _scale2int(x, dims, Dimensions.pairs2dims(scale...); verbose)
 @inline _scale2int(x, dims::DimTuple, scale::NamedTuple; verbose=true) = 
     _scale2int(x, dims, Dimensions.kw2dims(scale); verbose)
+@inline _scale2int(x, dims::DimTuple, scale::Dimension; verbose=true) = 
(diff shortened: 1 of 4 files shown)


tiemvanderdeure:
I was on it!

I also snuck a little change in to allow for scale by just a dimension to work - though it was a bug but on re-reading the docs we never promise it works. I can revert.

tiemvanderdeure:
There should be a better way than just a view, but I couldn't think of what it would be (other than implementing a whole new type with a custom `getindex`), and the performance is already quite good. I did some comparisons between the default eager disaggregation, and first lazily disaggregating and then collecting, and it is pretty much the same for in-memory arrays (I didn't try for on-disk arrays though).

rafaqz:
Totally, I don't think it's worth putting time into. I just mean view is one kind of lazy array we could use.

rafaqz:
Uggg Plots.jl has decided to return filenames from `savefig`. So we need semicolons after all those calls in the docs

PR #888, opened 2025-02-06
Breaking: drop the == dispatch on Rasters

### LINKED ISSUE #887: Remove the custom definition of `==` on 2 rasters?
Right now `==` on two Rasters has a custom definition, so that
```
da1 = rand(X(1:10))
da2 = rand(Y(1:10))

da1 == da2 # returns false
Raster(da1) == Raster(da2) # errors with DimensionMismatch
```

This is because of this definition
https://github.com/rafaqz/Rasters.jl/blob/177d89f5bc9c265b1ba002170551f7126f226983/src/array.jl#L69-L71

Which hasn't been touched for 4 years - should we just remove it on the next breaking release?

tiemvanderdeure:
The fallback is from DimensionalData

Let's see how much (if anything) this breaks

Closes https://github.com/rafaqz/Rasters.jl/issues/887


rafaqz:
Turns out this breaks a lot!

tiemvanderdeure:
> Turns out this breaks a lot!

Yeah which really makes sense because the method from DD actually checks if `dims(a) == dims(b)`.

With this PR `Raster(zeros(X(1:10))) == Raster(zeros(Y(1:10)))` just returns `false` instead of throwing an error. What I didn't realize is that comparing Rasters where dims have different loci will currently return `true`, now it is `false` - which I think is correct behaviour

tiemvanderdeure:
The other case where this is breaking is where dims are approximately equal so `comparedims` passes but `==` is false.

Importantly this happens often when writing a Raster and reading it back in, so this caused some test failures.

Makes me think we should have used `comparedims` in the method definition over in DD...

tiemvanderdeure:
I've changed the tests so they pass now, it didn't break all too many of them. 

I'm still not sure if it's a good or a bad thing that == would be false after writing a Raster and reading it back in. On the one hand it is a bit unintuitive, e.g. there was a case in the tests where all lookup values were equal, but the step was a floating point unit off, and so it is hard to understand why these are not equal. But on the other hand in many cases some information is lost when reading and writing - floating point error is just one example, in some cases the locus ends up being shifted, and Dates with regular intervals also lose that information. 

The current definition of `==` doesn't make a whole lot of sense, so if we really think it should return true when dimensions are within floating point, we could overwrite it here as `comparedims(Bool, a,b) && parent(a) == parent(b)` and then change it to match on the next breaking version of DD.

What do you think @rafaqz?

felixcremer:
One thing I found just now is that == is true when underlying data is the same, but the missingvalue is different. This is confusing, because the difference in the missing value will change the behaviour in mask for example. 
So I am wondering whether we would need a better method in DimensionalData which would also take these extra fields of a struct into account or we would have to still have a method in Rasters. 

I find it fair, that == is false after writing and reading the data when the loci or some dimension changes. I find it much more confusing when == gives true and then the behaviour is slightly different.

tiemvanderdeure:
Did this get automatically closed because breaking was deleted @rafaqz ?

rafaqz:
Oh yeah seems it does that

rafaqz:
Breaking was merged ;)

PR #892, opened 2025-02-09
Always wrap variables in `ModifiedDiskArray` even when it does nothing

### LINKED ISSUE #889: writing a lazy raster doesn't use chunks
MWE:

```julia
using Rasters, NCDatasets
ras1 = Rasters.create("ras1.nc", Float64, (X(1:100), Y(1:100)); force = true, chunks = (10, 10))
@time write("ras2.nc", ras1; force = true, chunks = (10,10))
```
is slow! `0.284892 seconds (1.14 M allocations: 41.458 MiB)`. I checked and it does call `setindex_disk!` for every index instead of once per chunk.

YAXArrays does manage to do this chunk-wise (although it's extremely slow because of a `GC.gc()`)

```julia
using YAXArrays, NetCDF
yax1 = open_dataset("ras1.nc")
@time savecube(yax1.unnamed, "ras2.nc", overwrite = true)
```

Something like this is similarly slow

```julia
open(ras1) do src
    open(Raster("ras2.nc"; lazy = true); write = true) do dest
        @time dest .= src
        nothing
    end
end
```

rafaqz:
We unfortunately can't rely on CommonDataModel.jl packages to implement the DiskArrays.jl interface properly, e.g. #889

But they at least have `eachchunk`/`haschunks`/`readblock!` etc.

So as a general strategy we will wrap them in a properly implemented disk array and prevent them from ever being exposed to dispatch methods other than `readblock!` etc.

Closes #889


rafaqz:
@felixcremer the zarr test file on bgi jena doesn't seem to exist anymore. Do you have a more reliable long term test file? Or can we put one somewhere? I've commented out the zarr tests for now but thats less than ideal, especially would be good to test these changes with zarr too

felixcremer:
The data should still be there and I can access them. They should not be removed as far as I know.

rafaqz:
Ok maybe it was just down or my conenction was bad at the time

PR #893, opened 2025-02-09
actually use the opened dataset in zonal

asinghvi17:


rafaqz:
ah yes, the merge conflict won this time

Issue #895, opened 2025-02-14
Cellarea crashes Julia on engineering crs


evetion:
```julia
using Rasters, Proj, ArchGDAL
r = Raster("https://github.com/Deltares/Geomorphometry.jl/releases/download/v0.6.0/saba.tif")
cellarea(r)
```

asinghvi17:
Can confirm...that seems bad

asinghvi17:
`cellarea(Planar(), r)` works as expected.  The problem is that this CRS has no geographic CRS backing it and so there is no way to go to lat/long.  We need to detect that in the main function and dispatch...

rafaqz:
Probably has to be Proj or GDAL to kill Julia?

asinghvi17:
Yes the stacktrace points to Proj

tiemvanderdeure:
This must be a Proj issue, right? Seems like it segfaults if it can't do a conversion and `always_xy = true`.

```julia
using Rasters, Proj, ArchGDAL
r = Raster("https://github.com/Deltares/Geomorphometry.jl/releases/download/v0.6.0/saba.tif"; lazy = true)

transform = Proj.Transformation(crs(r), EPSG(4326)) # errors
transform = Proj.Transformation(crs(r), EPSG(4326); always_xy = true) # segfaults
```

But also this CRS seems to be invalid? `ArchGDAL.importCRS(crs(r))` gives `ERROR: Failed to export this SRS to PROJ.4 format (Failure.)`

evetion:
It's a local engineering CRS. Should be valid, but you can't transform it. It should error though, not crash, and not sure if that's a bug at the Julia side of Rasters, Proj or PROJ itself.

rafaqz:
It's pretty hard for Rasters to cause a default, and really Proj should protect other packages from this

tiemvanderdeure:
I made an issue in Proj, think it should be handled there

rafaqz:
Closing as not a Rasters.jl issue

PR #896, opened 2025-02-14
experiment with best_manifold implementation here

asinghvi17:
seems we need this now, will move the code to GeometryOps / Proj / GFT / wherever later but want to at least get a working implementation here.

this is the way to solve #895

### DIFF
--- a/src/extensions.jl
+++ b/src/extensions.jl
@@ -207,8 +207,10 @@ cs = cellarea(myraster)
 ```
 $EXPERIMENTAL
 """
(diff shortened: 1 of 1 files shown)


asinghvi17:
It looks like we will have a problem with detecting things that don't have a backing ellipsoid - `proj_get_ellipsoid` segfaults on this CRS.

tiemvanderdeure:
I get the point here but this also seems quite fragile and blackboxy. We could also just throw a more informative errors that point users towards passing `Planar`

PR #902, opened 2025-02-19
Cdm backendexception

tiemvanderdeure:
On main `using Rasters; Raster("myraster.nc")` returns `ERROR: MethodError: no method matching checkfilename(::Rasters.NCDsource, ::String)`. I think the backendexception that is intended here was broken with the introduction of CDM?


rafaqz:
Yeah SafeTestSets is not safe for loading extensions unfortunately. Checks like that have to go at the start

Issue #909, opened 2025-02-25
`extract` returns the wrong `index`


tiemvanderdeure:
```MWE
ras = Raster(rand(X(1:10), Y(10:-1:1)))
ex = extract(ras, (1,10), index = true)
ex.index
```

This only happens for `ReverseOrdered` lookups. What happens internally is that  `_prepare_for_burning` changes the lookups. Similar to https://github.com/rafaqz/Rasters.jl/issues/876

tiemvanderdeure:
I'm very confused by why this ever passed tests though

rafaqz:
No tests with Reverse order, easy mistake to make

Issue #876, opened 2025-01-27
floating-point issues when `extract`ing the edge of a grid cell


tiemvanderdeure:
In the example below, we extracting from a raster defined by its edges, and in half the cases we don't get the correct cell. In lots of cases we also get `missing`, sometimes when extracting from the middle of the Raster!

```julia
using Rasters, Rasters.Lookups
xdim = 0:0.0416666665:1
ydim = 5:-0.0416666665:-5
ds = (X(xdim; sampling = Intervals(Start())), Y(ydim; sampling = Intervals(Start())))
ras = Raster(DimPoints(ds); dims = ds)
extr = extract(ras, DimPoints(ds), skipmissing = false)
extract_itself = getfield.(extr, 1) .== getfield.(extr, 2)
count(!, skipmissing(extract_itself)) # 3631
count(ismissing, extract_itself) # 150
length(extract_itself) # 6025
```

I encountered this in a case where I `Rasters.sample` some points from one raster and then `extract` them from another with identical dimensions. Which probably isn't too crazy of a use case.

I did a little bit of digging and what is going on here is that the Raster dimensions get converted from a range to a vector of values by `_prepare_for_burning`, and that allows for floating point issues to become a thing.

rafaqz:
Ahh damn. Does it work if we don't convert to vectors?

tiemvanderdeure:
I tried just taking `_prepare_for_burning` out that fixes this issue and tests pass. Why exactly is it needed? It is for performance?

rafaqz:
Yeah vectors are much faster. But I think there are other things in that method we need to keep, we can just remove the collector part

tiemvanderdeure:
Probably shiftlocus would also give issues - how badly do we need that? Otherwise we can keep it for polygons but not for points?

rafaqz:
Shiflocus is a no-op for points. But for intervals it would be a lot of work to handle rasterizing all variants

asinghvi17:
Hmm...wouldn't you just need `fast_index(locus, lookup, val)`? After that, the grid is the grid is the grid...

rafaqz:
Not sure what you mean there... Rasterization uses y values to make edges, x values during scans. At some point you need to check when you cross them. It also uses them in bresenham.

So there are multiple points you just have to get values out of twice precision ranges and put them in something faster to access.

tiemvanderdeure:
> Rasterization uses y values to make edges, x values during scans.

[...]

PR #911, opened 2025-02-28
extract fixes: return the correct index and avoid some floating point issues

tiemvanderdeure:
Closes https://github.com/rafaqz/Rasters.jl/issues/876 and closes https://github.com/rafaqz/Rasters.jl/issues/909

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -1,11 +1,12 @@
 # Object to hold extract keywords
-struct Extractor{T,P,K,N<:NamedTuple{K},ID,G,I,S,F}
+struct Extractor{T,P,K,N<:NamedTuple{K},ID,G,I,S,F,O<:Tuple{Order, Order}}
     names::N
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
Hm this does affect performance a little bit. Maybe we need another function barrier

```julia
using Rasters, Chairmarks
import GeoInterface as GI
line = GI.Line([(8.0, 0.0), (12.0, 4.0)])
lines = [line for i in 1:1_000]

dimz = (X(10.0:-1.0:9.0), Y(0.1:0.1:0.2))
rast = Raster(Union{Int,Missing}[3 4; 1 2], dimz; name=:test, missingval=missing)
@b extract(rast, lines; geometry=false, threaded=true, flatten=true) |> length
```

On main `351.500 μs (7631 allocs: 980.859 KiB)`
On this branch `507.300 μs (7719 allocs: 982.141 KiB)`

PR #914, opened 2025-03-07
Breaking: more consistent (dis)aggregation of lookups

### LINKED ISSUE #913: `disaggregate` does not preserve extent
Now that it's not possible to pass a locus to `disaggregate`, it always disaggregates at the Start, which returns some pretty unexpected dimensions (to me). I think it should be a priority to always return the same bounds when disaggregating - right now that doesn't happen.

```julia
using Rasters, Rasters.Lookups
ras = Raster(rand(X(Sampled(5:10:95; sampling = Intervals(Center())))))
ras2 = disaggregate(ras, 10)

extent(ras2) == extent(ras)
```

I thought this would be an easy fix, but it requires changing some of the internals and then everything gets messed up.

So maybe the solution is to `shiftlocus` internally or something. Or do should we make it possible to pass a locus again?

tiemvanderdeure:
closes https://github.com/rafaqz/Rasters.jl/issues/913


src/methods/aggregate.jl
```diff
@@ -272,12 +275,13 @@ function disaggregate(lookup::Lookup, scale)
...
-    if lookup isa AbstractSampled
+    newlookup = if lookup isa AbstractSampled
         sp = disaggregate(locus, span(lookup), intscale)
-        return rebuild(lookup; data=index, span=sp)
+        rebuild(lookup; data=index, span=sp)
     else
-        return rebuild(lookup; data=index)
+        rebuild(lookup; data=index)
     end
+    return maybeshiftlocus(loc, newlookup)
```
rafaqz, line 284: The double locus shift is a bit worrying, easy to introduce fp errors? Can we not explicitly create start/center/end lookups instead?

tiemvanderdeure: yeah it's not ideal - let's find a better solution

tiemvanderdeure:
Okay I gave it a go now with a more proper fix and added some tests to show what my goal is here.

Some of the tests fail and I'm not totally sure how to fix them yet, will look more at it later.

But it also seems the tests 1) still use some old (deprecated?) syntax (e.g. passing a locus to `disaggregate`), and 2) mostly test point Rasters, where I think most of the actual usage of (dis)aggregate will be on interval sampling. It's also much more intuitive what disaggregation of intervals should look like vs disaggregation of points.

rafaqz:
Sorry havent had time to look at this, did you figure it out?

tiemvanderdeure:
I figured something out, but if IIRC the thing I figured out would be breaking

rafaqz:
We can bundle it with the DD breaking change

tiemvanderdeure:
Now I remember - we need to choose if we still want to allow `Locus` to be passed in as the first argument. This still works (and is tested!), but is not documented.

tiemvanderdeure:
More specifically my idea here is that the clearest case for disaggregating is with interval sampling. We can visualize breaking each grid cell into smaller grid cells, and that's what I implemented here. Before this PR that would depend on the sampling.

It's less clear how disaggregation should work for points, which is where this locus argument came in - but again, at the moment that is not documented

src/methods/aggregate.jl
```diff
@@ -167,6 +166,7 @@ end
 function aggregate!(f, dst::AbstractRaster, src, scale;
     skipmissingval=false, skipmissing=skipmissingval, progress=true, verbose=true
 )
+    @show dims(dst)
```
felixcremer, line 169:
```suggestion
```

tiemvanderdeure:
At the moment the PR will just always disaggregate point from the center. If you think that is a reasonable way to go all we need to do is fix the tests and this PR is good to go. I'm away for the weekend but can do it on Monday

src/methods/aggregate.jl
```diff
@@ -107,19 +107,18 @@ aggregate(method, l::Lookup, scale::Colon) = aggregate(method, l, length(l))
 aggregate(method, l::Lookup, scale::Nothing) = aggregate(method, l, 1) 
 function aggregate(method, l::Lookup, scale::Int)
     intscale = _scale2int(Ag(), l, scale)
-    start, stop = _endpoints(method, l, intscale)
-    if issampled(l) && isordered(l)
-        newl = l[start:scale:stop]
-        sp = aggregate(method, span(l), scale)
-        return rebuild(newl; span=sp)
+    if issampled(l) && isordered(l) && isregular(l)
```
rafaqz, line 110: Isn't aggregating `Irregular` ok in some cases? 

Like its a bit wrong to do with mean but fine with sum?

src/methods/aggregate.jl
```diff
@@ -150,7 +149,7 @@ function aggregate!(loci::Tuple{Locus,Vararg}, dst::AbstractRaster, src, scale;
 )
     comparedims(dst, src; length=false)
     intscale = _scale2int(Ag(), dims(src), scale; verbose)
-    offsets = _agoffset.(loci, intscale)
+    offsets = ceil.(Int, _agoffset.(loci, (ForwardOrdered(),), intscale))
```
rafaqz, line 152: What does `ceil` do here? a comment would help

src/methods/aggregate.jl
tiemvanderdeure: This is in lookup aggregation. How should we aggregate irregular lookups?

src/methods/aggregate.jl
tiemvanderdeure: Had to add it because I replaced a `÷` with a `\` in `_agoffset` - the behaviour is unchanged. But I've added a comment

src/methods/aggregate.jl
rafaqz: Just like they were in the past... Start and end locus are especially simple. It's just dropping the other values.

Center is a bit more complicated, I guess the middle of the aggregated range?

tiemvanderdeure: I added an extra check for regular - so for irregular lookups it works basically like it used to.

tiemvanderdeure: Actually we might want the old behaviour for point lookups as well - I'm not sure

tiemvanderdeure:
I think this PR is good to go now. 

All the arithmetics are super annoying but just to show what this PR did, let me plot some rasters before/after (dis)aggregation.
Before this PR:
billede
After this PR:
billede

Code here is just:
```julia
using Rasters, GLMakie, Rasters.Lookups
ras = rand(X(Sampled(1:10; sampling=Intervals(Center()))), Y(Sampled(1:10; sampling=Intervals(Center())))) |> Raster
function plot_outline!(ax, ras; color, kw...)
    bnds = Rasters.intervalbounds(ras)
    bnds_A = tuple.(bnds[1], permutedims(bnds[2]))
    broadcast(bnds_A, color) do bnds, color
        (x, y) = bnds
        poly!(ax, [(x[1], y[1]), (x[2], y[1]), (x[2], y[2]), (x[1], y[2])]; color, kw...)
    end
end

rasag = aggregate(mean, ras, 2)
disag = disaggregate(rasag, 2)

fig = Figure(size = (1000, 500))
ax = Axis(fig[1,1])
plot_outline!(ax, ras, color=parent(ras), strokewidth=2, colorrange = extrema(ras), alpha = 0.5)
plot_outline!(ax,rasag, color=parent(rasag), strokewidth=5, colorrange = extrema(ras), alpha = 0.5)
ax = Axis(fig[1,2])
plot_outline!(ax, disag, color=parent(disag), strokewidth=2, colorrange = extrema(ras), alpha = 0.5)
plot_outline!(ax, rasag, color=parent(rasag), strokewidth=5, colorrange = extrema(ras), alpha = 0.5)
fig
```

rafaqz:
Would you say breaking change or bugffix?

tiemvanderdeure:
I'm not even sure. Some of the changes for point sampling might be actually breaking? But some of that behaviour was already deprecated

tiemvanderdeure:
But if there's a breaking release of DD on the way we might as well just add it to the breaking release here?

rafaqz:
Yeah, it makes sense I'm just not sure when it will all happen

tiemvanderdeure:
Ready to merge I think!

Issue #910, opened 2025-02-27
better ways to dodge floating point issues in `crop`


tiemvanderdeure:
I know we've discussed this but actually couldn't find an issue and it's pretty annoying so I'll make one (and apologies if I've overlooked an earlier issue).

I always run into cases like this where `crop` just doesn't return what I want because of floating-point error

Silly example where I can't even crop this Raster to itself:

```julia
A = Raster(rand(X(Sampled(0:(1/12):1, sampling = Intervals(Start())))))
A2 = crop(A; to = A)
size(A) == size(A2) # false
```

When writing a raster and reading it back in this happens all the time.

I know touches=true exists but often it doesn't work because then you get the problem in the other direction.

Should we add an `atol` keyword? Or an option to crop to cells where the center is covered?

rafaqz:
But where does atol actually get applied

tiemvanderdeure:
> But where does atol actually get applied

Would just padding the extent by `atol` before we pass it to `view` work?

I know it probably isn't trivial to solve this but I figured it'd be useful to have an issue open.

PR #919, opened 2025-03-10
Add `atol` keyword to crop

tiemvanderdeure:
Closes https://github.com/rafaqz/Rasters.jl/issues/910

Seems like a simple solution to an annoying problem to me. What do you think @rafaqz ?


rafaqz:
Yeah, ses pragmatic

tiemvanderdeure:
this one is good to go as far as I'm concerned. Unless someone has a better idea for how this could work.

PR #924, opened 2025-03-24
fix offset handling in extract

tiemvanderdeure:
Fixes 2 issues in extract:
- for linestring, the resize! just resized to the maximum length of the latest line segment, instead of the total - triggering out of bounds indexing errors
- for line and polygon burning, offset wasn't applied in `_prop_nt`, meaning the wrong values got extracted! I hope this got introduced by https://github.com/rafaqz/Rasters.jl/pull/911/, but I'm not sure. It evaded tests because we test on 2x2 rasters where the offset is always 0

### DIFF
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -306,7 +306,7 @@ end
     dims, offset = _template(A2, geom)
     dp = DimPoints(A2)
     function _length_callback(n)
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
Just a reminder we need this merged before the next patch version, otherwise extract will be very broken

PR #929, opened 2025-03-28
update versions

lazarusA:
- [x] latest DVPress changes.
- [x] fixed dead links
- [x] added Source doc string
- [x] added `mosaic(...; progress=false)`.


rafaqz:
Thanks!! @evetion maybe you can copy this too

lazarusA:
ok, I don't know why the plots from plots are not being saved. Locally, if I run the corresponding code it works! why plots? why? @rafaqz ?

BTW, is the only thing failing.

rafaqz:
Ugh yeah I hate it, I remember struggling with the file path... Is it easier if we make them all Makie?

lazarusA:
Paths are fine, the issue is that the `jldocstests` are not running when doing `make.jl`, and is only for `mosaic`. Don't know if a global setting is missing or something.


I could simply do these 2 in Makie then 😄 .

lazarusA:
all good now, will merge. But, someone else would need to fix this https://rafaqz.github.io/Rasters.jl/previews/PR929/tutorials/spatial_mean#Bonus:-Computing-spatial-means-across-dimensions later, not sure who did this part, @asinghvi17 or @tiemvanderdeure ? maybe.

asinghvi17:
That would be Tiem

PR #936, opened 2025-04-10
load CDM CRS if they are available in string form

asinghvi17:
we still cannot load CF CRS natively, but most writers will also write the CRS in string form, which we can read.  The key indicates the format.

### DIFF
--- a/src/sources/commondatamodel.jl
+++ b/src/sources/commondatamodel.jl
@@ -168,6 +168,10 @@ function _layers(ds::AbstractDataset, names, group)
 end
 
 function _dims(var::AbstractVariable{<:Any,N}, crs=nokw, mappedcrs=nokw) where N
+    if isnokw(crs)
+        # Get CRS by hook or by crook - start at the variable and work our way up
+        crs = _get_crs_by_hook_or_crook(CDM.dataset(var), var, CDM.attribs(var))
+    end
     dimnames = CDM.dimnames(var)
     ntuple(Val(N)) do i
         _cdmdim(CDM.dataset(var), dimnames[i], crs, mappedcrs)
@@ -339,6 +343,48 @@ function _cdmlookup(
     end
     return _cdmlookup(D, index, order, span, sampling, metadata, crs, mappedcrs)
 end
+
+function _get_crs_by_hook_or_crook(ds::AbstractDataset, var, attr)
+    # Get CRS by hook or by crook - start at the variable and work our way up
+    crs = nokw
+    if haskey(attr, "grid_mapping")
+        if haskey(ds, attr["grid_mapping"])
+            crs = _crs_from_cf_attr(CDM.attribs(ds[attr["grid_mapping"]]))
+            !isnokw(crs) && return crs
+        else
+            crs = _crs_from_cf_attr(attr)
+            !isnokw(crs) && return crs
+        end
+    elseif haskey(CDM.attribs(var), "grid_mapping") 
+        var_attr = CDM.attribs(var)
+        if haskey(ds, var_attr["grid_mapping"])
+            crs = _crs_from_cf_attr(CDM.attribs(ds[var_attr["grid_mapping"]]))
+            !isnokw(crs) && return crs
+        else
+            crs = _crs_from_cf_attr(var_attr)
+            !isnokw(crs) && return crs
+        end
+    else# nothing locally...check global attributes as a last resort
+        return _crs_from_cf_attr(attr)
+    end
+
+end
+
+function _crs_from_cf_attr(attr)
+    # we can't parse CF crs fully, BUT
+    # many CF implementations will include either `crs_wkt`, `spatial_epsg`, or `proj4string`
+    # which we do know the types of
+    if haskey(attr, "crs_wkt")
+        return ESRIWellKnownText(GeoFormatTypes.CRS(), attr["crs_wkt"])
+    elseif haskey(attr, "spatial_epsg")
+        return EPSG(parse(Int, attr["spatial_epsg"]))
+    elseif haskey(attr, "proj4string")
+        return ProjString(attr["proj4string"])
+    end
+    return nokw # in the worst case return nothing
+end
+
+
 # For X and Y use a Mapped <: AbstractSampled lookup
 function _cdmlookup(
     D::Type{<:Union{<:XDim,<:YDim}}, index, order::Order, span, sampling, metadata, crs, mappedcrs


asinghvi17:
Looks like I did something fishy here that leads CRS to return a bool

alex-s-gardner:
When I test on my end I still get an `nothing` returned

```
using Rasters
import NCDatasets

paths2granule = "https://hyp3-jhk-sandbox-contentbucket-wxvu5zy6v7d6.s3.us-west-2.amazonaws.com/100f0e79-d569-42b9-8a20-3c3384b34dee/S1B_IW_SLC__1SSH_20181227T194309_20181227T194320_014232_01A75C_12D4_X_S1A_IW_SLC__1SSH_20190102T194350_20190102T194401_025303_02CCA3_D38F_G0120V02_P099.nc"

local_path = joinpath(tempdir(), splitpath(paths2granule)[end])

isfile(local_path) || Downloads.download(paths2granule, local_path)

rs = RasterStack(local_path; name=[:v])

Rasters.crs(rs)
```

asinghvi17:
So it looks like the RasterStack constructor doesn't follow this pipeline.

```julia-repl
julia> crs(Raster(local_path; name=:v))
ESRIWellKnownText with CRS mode: PROJCS["WGS 84 / NSIDC Sea Ice Polar Stereographic North",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4326"]],PROJECTION["Polar_Stereographic"],PARAMETER["latitude_of_origin",70],PARAMETER["central_meridian",-45],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Easting",SOUTH],AXIS["Northing",SOUTH],AUTHORITY["EPSG","3413"]]
```

src/sources/commondatamodel.jl
```diff
@@ -339,6 +343,48 @@ function _cdmlookup(
...
+    end
+
+end
+
+function _crs_from_cf_attr(attr)
+    # we can't parse CF crs fully, BUT
+    # many CF implementations will include either `crs_wkt`, `spatial_epsg`, or `proj4string`
+    # which we do know the types of
+    if haskey(attr, "crs_wkt")
+        return ESRIWellKnownText(GeoFormatTypes.CRS(), attr["crs_wkt"])
```
rafaqz, line 378: An reason this is `ESRIWellKnown` text and not the normal one?

asinghvi17: Not particularly, no
```suggestion
        return WellKnownText2(GeoFormatTypes.CRS(), attr["crs_wkt"])

File: src/sources/commondatamodel.jl [hunk 2]
```diff
@@ -339,6 +343,48 @@ function _cdmlookup(
     end
     return _cdmlookup(D, index, order, span, sampling, metadata, crs, mappedcrs)
 end
+
+function _get_crs_by_hook_or_crook(ds::AbstractDataset, var, attr)
```
rafaqz, line 347: One problem with this is we are assuming which dimensions crs should attach to instead of reading the explicit link to x/y etc dimensions in the grid mapping.

I think we need to return something like ("lat", "lon") => crs so we can match them properly when we make dimensions.

asinghvi17: > explicit link to x/y etc dimensions in the grid mapping.

Isn't that in the actual x/y dims though?  This is reading from the variable `var` primarily...

rafaqz: Yeah but it still lists which dimensions of the variable the crs relates to, and we should probably keep that information even though nearly always its obvious which those should be.

asinghvi17: We don't have anywhere to store that in Projected though.  And we already do that when choosing the dims for the variable right?

Or is that just following the order of definition?

That sounds too complicated and like too much of a rework to go in this PR though, if we just want CRS support.  It can come later but I have a feeling that no one actually screws around with dimensions that much.

rafaqz: Everything else about dimensions come just from the dimension variable and other variables (like bounds) that it explicitly links to.

This is quite different - its putting some metadata from another variable into the dimensions - so we need to make a link from variable to its dimensions - but some of those are the crs dimensions, some are other dimensions.

It is actually explicitly specified which dimensions the crs relates to, so instead of guessing the link we should use that information.

asinghvi17: So I'm not sure what exactly you're talking about here, we may be talking past each other.  Here are all the relevant attributes from a file.  I've removed the wkt crs and other variables to focus on the one thing.

The grid mapping has no mention of axes whatsoever.  The variable also only specifies dimensions; it's the individual axis vars that tell you what dimension they are.  It's perfectly acceptable for a variable to be named `x` but actually represent dimension `y`, if confusing.

If you're saying that we don't currently check the "dimensions" bit, yes we should do that.  If we do check that and you want to check that the CRS actually matches the dimension, we can't do that without Proj which knows how to decompose the CRS.

Or do you mean the "T:coordinates" field here?

```
dimensions:
  rlon = 128 ;
  rlat = 64 ;
  lev = 18 ;
variables:
  float T(lev,rlat,rlon) ;
    T:long_name = "temperature" ;
    T:units = "K" ;
    T:coordinates = "lon lat" ;
    T:grid_mapping = "rotated_pole" ;
  char rotated_pole ;
    rotated_pole:grid_mapping_name = "rotated_latitude_longitude" ;
    rotated_pole:grid_north_pole_latitude = 32.5 ;
    rotated_pole:grid_north_pole_longitude = 170. ;
  float rlon(rlon) ;
    rlon:long_name = "longitude in rotated pole grid" ;
    rlon:units = "degrees" ;
    rlon:standard_name = "grid_longitude";
  float rlat(rlat) ;
    rlat:long_name = "latitude in rotated pole grid" ;
    rlat:units = "degrees" ;
    rlat:standard_name = "grid_latitude";
  float lev(lev) ;
    lev:long_name = "pressure level" ;
    lev:units = "hPa" ;
  float lon(rlat,rlon) ;
    lon:long_name = "longitude" ;
    lon:units = "degrees_east" ;
  float lat(rlat,rlon) ;
    lat:long_name = "latitude" ;
    lat:units = "degrees_north" ;
```

If so then yeah, we should definitely check that.

If it's none of those then I have no idea where you're going with this.

```
mapping
    Attributes:
     grid_mapping_name    = polar_stereographic
     straight_vertical_longitude_from_pole = -45.0
     false_easting        = 0.0
     false_northing       = 0.0
     latitude_of_projection_origin = 90.0
     latitude_of_origin   = 70.0
     semi_major_axis      = 6.378137e6
     scale_factor_at_projection_origin = 1
     inverse_flattening   = 298.257223563
     proj4text            = +proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs
     spatial_epsg         = 3413
     GeoTransform         = -60787.5 120.0 0 -1340932.5 0 -120.0

  vx   (920 × 960)
    Datatype:    Union{Missing, Int16} (Int16)
    Dimensions:  x × y
    Attributes:
     _FillValue           = -32767
     standard_name        = land_ice_surface_x_velocity
     description          = velocity component in x direction from radar range and azimuth measurements
     units                = meter/year
     grid_mapping         = mapping
     error                = 7.8
     error_description    = best estimate of x_velocity error: vx_error is populated according to the approach used for the velocity bias correction as indicated in "stable_shift_flag"
     error_stationary     = 7.8
     error_stationary_description = RMSE over stable surfaces, stationary or slow-flowing surfaces with velocity < 15 meter/year identified from an external mask
     error_modeled        = 44.6
     error_modeled_description = 1-sigma error calculated using a modeled error-dt relationship
     error_slow           = 7.9
     error_slow_description = RMSE over slowest 25% of retrieved velocities
     stable_shift         = -7.8
     stable_shift_flag    = 1
     stable_shift_flag_description = flag for applying velocity bias correction: 0 = no correction; 1 = correction from overlapping stable surface mask (stationary or slow-flowing surfaces with velocity < 15 meter/year)(top priority); 2 = correction from slowest 25% of overlapping velocities (second priority)
     stable_shift_stationary = -7.8
     stable_count_stationary = 533584
     stable_shift_slow    = -6.8
     stable_count_slow    = 170823

  x   (920)
    Datatype:    Union{Missing, Float64} (Float64)
    Dimensions:  x
    Attributes:
     _FillValue           = NaN
     standard_name        = projection_x_coordinate
     description          = x coordinate of projection
     units                = m

  y   (960)
    Datatype:    Union{Missing, Float64} (Float64)
    Dimensions:  y
    Attributes:
     _FillValue           = NaN
     standard_name        = projection_y_coordinate
     description          = y coordinate of projection
     units                = m
```

rafaqz: I mean, I didn't know exactly what its called or where they are stored because I don't have a file like this locally. But there _has_ to be a variable link like that, netcdf/CF standards just work like this.

So yes this one seems like what I mean, but a bit more complicated due to linking the rotation:
```
 T:grid_mapping = "rotated_pole" ;
 T:coordinates = "lon lat" ;
```

In the one below you have this:
```
    Dimensions:  x × y
    Attributes:
     _FillValue           = -32767
     standard_name        = land_ice_surface_x_velocity
     description          = velocity component in x direction from radar range and azimuth measurements
     units                = meter/year
     grid_mapping         = mapping
 ```

Which links `x` and `y` as the dimensions of `grid_mapping`. 

But, I'm not sure what happens if this variable has more dimensions - how is it clarified which ones relate to `grid_mapping`.

rafaqz: Like clearly its `x` and `y` but they can kind of be named anything at all

rafaqz: Currently Rasters.jl is making that "obvious" assumption that xish and yish variables are where the crs should go, if you pass in crs manually. But thats mostly a hack to get around not easily being able to get the crs on most cases.

But when we have explicit links between `grid_mapping` and dimensions we should use them

asinghvi17: Hmm it seems like this is more at the variable / dimension level?  The CRS itself holds no metadata about what variables use it.  So we could, potentially, check the dimension attributes for "y coordinate of projection" or "longitude" etc, to make that link.  Could you point me to where in the code the dimension discovery happens?

rafaqz: This is where we link dimensions to variables

https://github.com/rafaqz/Rasters.jl/blob/main/src/sources/commondatamodel.jl#L184-L189

CommonDataModel provides `CDM.dimnames(var)` that gives you the string names of the dimension variables connected to any variable.

rafaqz: From the standard:

"An application can determine that x and y are the projection coordinates by recognizing the standard names projection_x_coordinate and projection_y_coordinate"

rafaqz: This will link arbitrary dimension names to spatial x/y

asinghvi17:
Should we get this in for now, if we can fix the issues with returning Bools, and then get the CF projection stuff out with the dimension linking whenever it's ready?

CF also has different names depending on geodetic/projected CRS for dimensions, which is another pain...

rafaqz:
No... Because then it will be breaking to do the dimensions linking ;)

asinghvi17:
Wouldn't that be breaking now as well?  Or is it not breaking because we don't return anything now?

rafaqz:
Yeah maybe they're both breaking

src/sources/commondatamodel.jl
asinghvi17: reading this code again I finally get what you mean :D - we just need to make the dimension type lookup do something different from what it currently does (name pattern matching)

PR #937, opened 2025-04-11
allow Raster(WorldClim{Elevation}) without layer argument

tiemvanderdeure:
Does what it says on the tin

closes https://github.com/EcoJulia/RasterDataSources.jl/issues/77

It's not super pretty to special-case this one datasource, but I think it's the only one that doesn't have a layer argument, so I guess it's okay.

### DIFF
--- a/ext/RastersRasterDataSourcesExt/constructors.jl
+++ b/ext/RastersRasterDataSourcesExt/constructors.jl
@@ -28,6 +28,7 @@ function RA.Raster(T::Type{<:RDS.RasterDataSource}, layer::Union{Symbol,Int};
     filename = getraster(T, layer; rds_kw...)
     Raster(filename; name=RDS.layerkeys(T, layer), crs, ra_kw...)
 end
+RA.Raster(T::Type{WorldClim{Elevation}}; kw...) = Raster(T, :elev; kw...)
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
Hmm why is the documentation failing here...

tiemvanderdeure:
I can't figure out how to build locally on windows, but I really don't see how this 1-line PR could have broken any code, and I just tried to make this .mp4 locally on this branch and it works fine

lazarusA:
What's the issue with building locally? Follow the suggestions when it fails. There are some warnings if you are on windows 😬.

tiemvanderdeure:
Never got npm to work. But now I looked at it again and maybe my university just blocks it? I'm not sure. In any case, the build files are all there, also the rplot.mp4, so that seems to work like it should?

lazarusA:
sometimes, CI fails due to a timeout when downloading the datasets. Triggering it again 😄 . Let's see.

tiemvanderdeure:
Seems to have worked this time!

lazarusA:
its fine now 😄 .  This was the issue, on first run, the `jldoctest` will fail, when running locally a bunch of these will pop out!

Screenshot 2025-04-11 at 20 10 15

(not sure how to deal with this in the future, but I leave this here for reference).

rafaqz:
Oh no precompile in the doctests 😭

Maybe there is some way to hide the output

PR #940, opened 2025-04-14
fix variable name detection for CommonDataModel sources

### LINKED ISSUE #912: Import a monthly-based NCDataset into a Raster
Following [this Discourse thread](https://discourse.julialang.org/t/how-to-load-a-raster-of-a-nc-file-with-multiple-monthly-bands/126487/3)

I have "chelsa_monthly_precipitations_2014.nc" file, a NC datasets with monthly precipitation data and can't load it into a vector of `Raster` or a `RasterStack` object.

This is how it appears on QGIS:

Image

The problems is that it doesn't have a "variable" name I can use passing the "name" argument to the `Raster` constructor.

I can load it using `NCDatasets.NCDataset(ncpath)`, so perhaps makling a `Raster` constructor from a `NCDataset` could be an option ??

rafaqz:
Needs some tests still, with a file that will otherwise load incorrectly.

Fixes #912

### DIFF
--- a/src/sources/commondatamodel.jl
+++ b/src/sources/commondatamodel.jl
@@ -117,41 +117,32 @@ haslayers(::CDMsource) = true
 defaultcrs(::CDMsource) = EPSG(4326)
 defaultmappedcrs(::CDMsource) = EPSG(4326)
 
-function _nondimnames(ds)
+function _varnames(ds)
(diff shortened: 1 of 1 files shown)


rafaqz:
This is reworked in #965

PR #950, opened 2025-04-22
make zonal open stacks of length < 32

asinghvi17:
not sure if we want this size or not but that's easy to configure

### DIFF
--- a/src/methods/zonal.jl
+++ b/src/methods/zonal.jl
@@ -18,10 +18,10 @@ These can be used when `of` is or contains (a) GeoInterface.jl compatible object
 - `boundary`: for polygons, include pixels where the `:center` is inside the polygon,
     where the line `:touches` the pixel, or that are completely `:inside` inside the polygon.
     The default is `:center`.
-- `progress`: show a progress bar, `true` by default, `false` to hide..
-- `skipmissing`: wether to apply `f` to the result of `skipmissing(A)` or not. If `true`
+- `progress`: show a progress bar, `true` by default, `false` to hide.
+- `skipmissing`: whether to apply `f` to the result of `skipmissing(A)` or not. If `true`
     `f` will be passed an iterator over the values, which loses all spatial information.
-    if `false` `f` will be passes a masked `Raster` or `RasterStack`, and will be responsible
+    if `false` `f` will be passed a masked `Raster` or `RasterStack`, and will be responsible
     for handling missing values itself. The default value is `true`.
 
 # Example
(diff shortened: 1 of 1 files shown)


asinghvi17:
Probably needs some form of test as well?>

rafaqz:
Hmm I'm going to change how open works pretty soon (non closure version), so we can open more like 300

Yes nc it's just one object so the number of layers doesn't matter as much.

But let's wait and do it properly. We also need to fix the type stability of multy layer stacks, the scaling isn't good currently

asinghvi17:
Well some code like this will be needed anyway right?

Or do we actually want to change the way this works, and "chunk" the files into groups that we open one at a time?  Horrific I know, but if it's for that sweet sweet performance....

src/methods/zonal.jl
```diff
@@ -64,7 +64,13 @@ function zonal(f, x::RasterStack; of, skipmissing=true, kw...)
     # TODO: open currently doesn't work so well for large rasterstacks,
     # we need to fix that before we can go back to this being a single method
     # on `RasterStackOrArray`.
-    _zonal(f, _prepare_for_burning(x), of; skipmissing, kw...)
+    return if length(x) < 32
```
rafaqz, line 67: `length` is not the number of layers, but length like an array, or how many rows the dataframe would have ;)

You want `length(values(x))` or maybe the simplest for the compiler is actually `lenngth(layerdims(x))`

rafaqz:
I will try and just write it properly soon. Maybe we can merge this as a stopgap but not worth doing anything complicated. Have you tried it with 32 layers already? How is the compile time?

src/methods/zonal.jl
asinghvi17:
```suggestion
    return if length(layerdims(x)) < 32
```

asinghvi17:
Haven't tried with 32 layers but IIRC that's the recommended limit for static arrays, so I figure it's probably fine here too

rafaqz:
Different problem here... It's 32 deep closures!

src/methods/zonal.jl
```diff
@@ -64,7 +64,13 @@ function zonal(f, x::RasterStack; of, skipmissing=true, kw...)
     # TODO: open currently doesn't work so well for large rasterstacks,
     # we need to fix that before we can go back to this being a single method
     # on `RasterStackOrArray`.
-    _zonal(f, _prepare_for_burning(x), of; skipmissing, kw...)
+    return if length(layerdims(x)) < 32
```
asinghvi17, line 67:
```suggestion
    return if length(names(x)) < 5
```

asinghvi17:
Ah yeah it would do that...in that case the limit is 5 ish I believe

PR #951, opened 2025-04-22
Fix outbounds line burn

### LINKED ISSUE #923: Line burning is incorrect at edges
MWE
```julia
using GLMakie, Rasters
dimz = (X(10.0:-1:9.0), Y(0.1:0.1:0.2))
rast = Raster(Union{Int,Missing}[3 4; 1 2], dimz; name=:test, missingval=missing)
linestring = GI.LineString([(8.0, 0.0), (9.5, 0.0), (10.0, 0.4)])

# Let's see what this looks like
fig, ax, pl = plot(rast)
plot!(ax, linestring)

extract(rast, linestring) # extracts 3 grid cells

rast2 = set(rast; X = Intervals(Center()), Y = Intervals(Center()))
extract(rast2, linestring) # still extracts 3 cells

rast3 = extend(rast; to = Extent(X = (5, 15), Y = (-1, 1)))
extract(rast3, linestring, skipmissing = true) # extracts correctly

rasterize(linestring; to = rast, missingval = false, fill = true) |> count # 3
crop(rasterize(linestring; to = rast3, missingval = false, fill = true); to = rast2) |> count # 2
```

Image

When extracting or rasterizing this line from this raster, line burning suggests that it intersects on 3 grid cells. Plotting would suggest it is 2.

After extending the raster, it gives different (apparently more correct) results. Sampling doesn't change anything, so I guess this is something to do with the initial grid cell for line extraction?

rafaqz:
Closes #923

@tiemvanderdeure this is just your example as tests and trunc => floor

### DIFF
--- a/test/extract.jl
+++ b/test/extract.jl
@@ -211,12 +211,10 @@ end
     # Single LineString
     @test extract(rast, linestring) isa Vector{T}
     @test all(extract(rast_m, linestring) .=== T[
-        (geometry = (9.0, 0.1), test = 1)
         (geometry = (10.0, 0.1), test = 3)
         (geometry = (10.0, 0.2), test = missing)
     ])
     @test all(extract(rast_m, linestring; skipmissing=true) .=== Tsm[
-        (geometry = (9.0, 0.1), test = 1)
         (geometry = (10.0, 0.1), test = 3)
     ])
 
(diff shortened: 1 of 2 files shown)


rafaqz:
those tests were so embarrassing how did I not see that was wrong

tiemvanderdeure:
I guess it's only obvious to see that it is wrong once you plot it.

That was a surprisingly easy fix!

rafaqz:
I think I did plot it but our plotting in Makie was unreliable at the time, like half pixel out

Issue #952, opened 2025-04-24
spatial_mean, bonus section is broken.


lazarusA:
See:

https://rafaqz.github.io/Rasters.jl/stable/tutorials/spatial_mean#Bonus:-Computing-spatial-means-across-dimensions

tiemvanderdeure:
I'll take a look. My guess is it has to be related to downloading the data.

lazarusA:
yes, it would be nice to remove this line: 

https://github.com/rafaqz/Rasters.jl/blob/7574a8f14802f39b8afe43ddbb54844e90836994/docs/make.jl#L49 

well, set it to `false`.  Being more strict, so that we breaking things as less that possible, and if so, catch any minor break.

tiemvanderdeure:
Okay so if I set that to false and run locally I should actually get an error so I can fix this, right?

lazarusA:
your will get errors, for sure 😄 .

tiemvanderdeure:
Fixing all of those errors might be a little bit overambitious. Is there a way to build just this markdown file?

tiemvanderdeure:
Actually managed to build docs with warnonly=false locally. The only errors I ran into were printing of precompilation in docstrings. But just importing those packages in make.jl should fix that.

tiemvanderdeure:
Just checked and it still has no outputs on the docs page. I seem to be able to build locally. Can any of you replicate locally?

PR #954, opened 2025-04-24
Error on failures in doc build

tiemvanderdeure:
As suggested by @lazarusA in https://github.com/rafaqz/Rasters.jl/issues/952#issuecomment-2826725875.

Let's see what happens?


tiemvanderdeure:
Seems to work!

asinghvi17:
Ah I wish we could have CI status report false but the docs still deploy...

PR #961, opened 2025-05-03
Make JET happier

rafaqz:
Like DD we need better coverage.

A few of these are bugs in untested code, and the rest just removing edge cases that probably cant get hit anyway, but JET and the compiler still have problems with them.

### DIFF
--- a/src/create.jl
+++ b/src/create.jl
@@ -150,6 +150,9 @@ end
 # parent as the parent type so e.g. CuArray will propagate
 create(filename::Union{AbstractString,Nothing}, T::Union{Type,TypeNamedTuple}, A::AbstractRaster; kw...) =
     create(filename, T, dims(A); parent=parent(A), kw...)
+# Create RasterStack from filename, Type and dims
+create(filename::Union{AbstractString,Nothing}, T::Type, st::AbstractRasterStack{K}; kw...) where K =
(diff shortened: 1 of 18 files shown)


src/create.jl
```diff
@@ -187,13 +190,13 @@ function create(filename::Union{AbstractString,Nothing}, T::Union{Type,NamedTupl
     reverse_y=nokw,
     kw...
 )
-    dims = _extent2dims(extent; size, res, crs, sampling)
-    dims = if reverse_y isa Bool && reverse_y && hasdim(ds, Y())
-        DD.setdims(ds, reverse(DD.dims(ds, Y())))
+    dims1 = _extent2dims(extent; size, res, crs, sampling)
+    dims2 = if reverse_y isa Bool && reverse_y && hasdim(dims1, Y())
```
asinghvi17, line 194: istrue could be good here

src/methods/burning/allocs.jl
```diff
@@ -19,7 +20,7 @@ function _burning_allocs(x;
 ) 
     if threaded
         if isnothing(x)
-            [Allocs(nothing) for _ in 1:nthreads]
```
asinghvi17, line 22: What prompted this?

rafaqz: The nothing arg isn't needed now.

It worked before I'm just making it consistent with another change

PR #962, opened 2025-05-04
fix _name_or_firstname

asinghvi17:


rafaqz:
Yeah, this is all changing... Will push soon

PR #965, opened 2025-05-06
Cdm overhaul

asinghvi17:


asinghvi17:
@rafaqz is this ready to merge into the geometry lookup branch from your end?

asinghvi17:
Yea that works too

rafaqz:
35 CF examples now! - all loading but not all in the optimal way. Rotations are still missing, and I'm not sure how to handle these lookups with a center point and an explicit polygon boundary.

With a regular lookup `At` would select the center point, and `Contains` would select anything in the cell (we should have used `Coveredby` for that, need to DE-I9M-ize DD asap)

So for this it could do the same thing? and maybe for all geometry lookups? `At` gets the centroid/representative point, and `Contains` or whatever checks if the point is in the area.

(I'm thinking we might need a DE19M.jl or DE19MPredicates.jl package to hold all of these wrapper types, then we can use them everywhere, like `Extents.predicate(Contains(), a, b)` and we have the predicate separated from the context.)

asinghvi17:
> At gets the centroid/representative point, and Contains or whatever checks if the point is in the area.

I don't know about that....it seems pretty useless, the representative point must be within the geometry at any rate

asinghvi17:
> DE19MPredicates.jl

That would be nice...but would conflict with GeometryOps operations when we add those.  Plus Rasters has Touches which is not DE9IM touches...

rafaqz:
Yeah swapping DD touches and covers is where the idea can from. I don't think it will conflict, are there types in GO? We can just use them there too?

asinghvi17:
> Haha then we can select with geometries omg

We already can ;)

> are there types in GO

No but there will be.  And they have to subtype `GOCore.Operation`.  That's so that they fit in the whole declarative API, if not then it's not really feasible...

One thing we can do is that DE9IMPreds.jl depends on GOCore, and Extents, and actually fills in the Extents.predicate method for these predicates.  But then DD depends on GOCore, not sure if we want that.

rafaqz:
> it seems pretty useless

Maybe, but I'm not sure what else `At` would do, probably error.

rafaqz:
> We already can ;)

But like, with all the predicates even on a regular raster

asinghvi17:
The problem there is again the names will clash

asinghvi17:
> But like, with all the predicates even on a regular raster

what does that mean?  If I say Intersects(geom), then it can only give me either the bbox or a mergedlookup raster, neither of which are particularly nice.

How do we specify that the raster should be masked?  Maybe a new MaskingIndexSet type?

rafaqz:
I don't get why we need both objects, the idea was to use the same thing rather than clash...

asinghvi17:
I think we're thinking of two different objects.  I'm thinking of the GO Operations interface, i.e.

```julia
abstract type GeometricPredicate <: Operation end

struct Contains{Alg{M}, Exact, ...} <: GeometricPredicate end

nargs(::Contains) = 2
application_level(::Contains) = GO.TraitTarget{GI.AbstractGeometryTrait}()

function (::Contains)(::GI.PointTrait, p1, ::GI.PointTrait, p2)
end
```

or something like this

***

My understanding is that you are thinking of more of a DD Dimension or Selector like struct?

The issue that I have is that `GO.Contains() != DE9IM.Contains()`, which will inevitably lead to name clashes basically everywhere.  So we should figure out how to mitigate that.  Maybe we just say that predicates are not GO operations?  Not sure what the best solution is here.  But this is the issue in my mind.

I don't know how the Operation interface will turn out - maybe we can make it semi trait based.  Or maybe we don't end up doing it at all?  It's kind of tricky.

rafaqz:
Were not ;)

We can put all those details in another object that goes inside the predicate wrapper.

Then they can be shared. 

DD uses Dimension wrappers for a lot of unrelated things in exactly this way. You can even pass keywords to a Dimension and they can get found and constructed later depending on the context, e.g. `X(data; sampling=Intervals())` works.

asinghvi17:
This apparently points to my vectordatacubes branch right now.  Should this PR be closed and we open a new one to point to main?

PR #968, opened 2025-05-08
fix OpenStack calls

tiemvanderdeure:
`open(stack)` seems to be broken on main.

Needs test but not sure which files to use?


src/stack.jl
```diff
@@ -527,8 +527,8 @@ function DD.modify(f, s::AbstractRasterStack{<:FileStack{<:Any,K}}) where K
 end
 
 # Open a single file stack
-function Base.open(f::Function, st::AbstractRasterStack{K,T,<:Any,<:FileStack{X}}; kw...) where {X,K,T}
-    ost = OpenStack{X,K,T}(parent(st))
+function Base.open(f::Function, st::AbstractRasterStack{K}; kw...) where K
```
asinghvi17, line 530: Lol!  I didnt realize this was how it happens.

Can't we just define

```julia
Base.open(st::AbstractRasterStack{K}; kw...) = OpenStack(parent(st))
Base.close(st::AbstractRasterStack{K}; kw...) = somehow_flatten_and_invoke_on_openstack(close, st) # maybe replace with filearray
```

and that's `lazy=:open`?

It looks like the multifile implementation is recursive but we could figure that out.

rafaqz: Yeah, but that will be part of doing that properly everywhere?

rafaqz: These are the easy stacks where it's one dataset, often it's not

rafaqz:
Probably test in stack.jl ?

src/stack.jl
rafaqz: And fwiw `lazy=:open` is confusing syntax, it has to be something else

asinghvi17: Something else is fine, I don't even mind single-arg `Rasters.open(ras_or_filename)`

rafaqz: The reason not to only allow `open(rast)` is it needs a double open. 

`Raster(fn; open=true)` would open to get metadata and just stay open.

PR #974, opened 2025-05-14
Fix for new DV

asinghvi17:


docs/Project.toml
```diff
@@ -32,3 +32,4 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 [sources]
 SpeciesDistributionModels = {url = "https://github.com/tiemvanderdeure/SpeciesDistributionModels.jl/"}
 Rasters = { path = ".." }
+DocumenterVitepress = { url = "https://github.com/LuxDL/DocumenterVitepress.jl", rev = "as/fix-no-config" }
```
asinghvi17, line 35:
```suggestion
```

alex-s-gardner:
Can this me merged? dev docs are still broken

docs/make.jl
```diff
@@ -43,17 +43,16 @@ makedocs(
         devbranch = "main",
         devurl = "dev";
     ),
-    draft = false,
     source = "src",
     build = "build",
     warnonly=true,
+    draft = true
```
asinghvi17, line 49:
```suggestion
```

asinghvi17:
ci is still red, but let's see if this fixes it...there was a documentervitpress bug that only happens if draft=true

lazarusA:
let's also do only `:breaking` please! we don't want to build every minor version.

asinghvi17:
Wasn't minor the default?  I personally don't have a problem with building every minor version

lazarusA:
almost, but we still need this line:

https://github.com/LuxDL/DocumenterVitepress.jl/blob/432ffea517afa05424124801c43abd36ba1dbf32/docs/src/.vitepress/config.mts#L46

PR #987, opened 2025-06-27
WIP: support attribute tables

tiemvanderdeure:
Implements a new modification for raster attribute tables. In this standard, the .tif contains integers which correspond to a row in an attribute table. 

It's a pretty obscure standard. One example I've seen have integer values 0,1,2,3, and the other has 0, 12,13,14 for some reason. One has a column called VALUE and the other Value.

The idea here is to store a dict in a AbstractModifications and then just perform a lookup within readblock!

GDAL reads some of these tables but not all and then can have different file extension names, so `_tablemod` I still need to implement

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -365,8 +365,17 @@ function _raster(ds;
     data_out, dims_out, metadata_out, missingval_out = _open(source, ds; name=name1, group, mod=nothing) do var
         metadata_out = isnokw(metadata) ? _metadata(var) : metadata
         missingval_out = _read_missingval_pair(var, metadata_out, missingval)
+
         # Generate mod for scaling
-        mod = isnokw(mod) ? _mod(eltype(var), metadata_out, missingval_out; scaled, coerce) : mod
+        mod = if isnokw(mod) 
+            if !raw && isfile(filename .* ".vat.dbf") # todo: how to detect a RAT/VAT?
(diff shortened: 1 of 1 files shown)


rafaqz:
Looks great!

src/array.jl
```diff
@@ -365,8 +365,17 @@ function _raster(ds;
     data_out, dims_out, metadata_out, missingval_out = _open(source, ds; name=name1, group, mod=nothing) do var
         metadata_out = isnokw(metadata) ? _metadata(var) : metadata
         missingval_out = _read_missingval_pair(var, metadata_out, missingval)
+
         # Generate mod for scaling
-        mod = isnokw(mod) ? _mod(eltype(var), metadata_out, missingval_out; scaled, coerce) : mod
+        mod = if isnokw(mod) 
+            if !raw && isfile(filename .* ".vat.dbf") # todo: how to detect a RAT/VAT?
```
asinghvi17, line 371: We should make this be able to pick up ArchGDAL rats

asinghvi17: Because those have .tif.rat or something which is a whole different standard...

tiemvanderdeure: Yeah I have to think about how to check for all of that. I'm not sure if there are more formats than just those two? And if different file extensions than .tif could have a RAT or something similar?

PR #1017, opened 2025-09-13
improve disaggregate performance

### LINKED ISSUE #1016: `disaggregate` is type unstable
`disaggregate` allocates for Union eltypes

```
using Rasters
r = Raster(rand(X(1:1000), Y(1:1000)))
r2 = replace_missing(r)
@time disaggregate(r, 2)
@time disaggregate(r2, 2)
```

Timings (after compilation):   
```
0.049012 seconds (28 allocations: 30.519 MiB)
  1.054906 seconds (4.00 M allocations: 141.149 MiB, 5.11% gc time)
```

Profile:

Image

tiemvanderdeure:
closes #1016


tiemvanderdeure:
For some reason I cannot reproduce the actual type instability this was supposed to solve, BUT:

The test this ran into was actually just wrong. It multiplied a UInt8 array by 2, then wrote to a UInt8 array. For some reason this does not trigger an InextactError, even though it should. The actual change (fill! instead of .=) actually decreases performance on lazy rasters. So not totally sure what to do with this PR :upside_down_face: 

Probably we keep the change to the test and drop the change to disaggregate?

```
dst = Rasters.create(tempname() * ".tif", UInt8,Extent(X=(0,100), Y=(0,100)); res = (1.0, 1.0))
# no error
open(dst; write = true) do d
    d .= 10000
end
# error
open(dst; write = true) do d
    fill!(d, 10000)
end
# error
read(dst) .= 10000
```

tiemvanderdeure:
```
ERROR: LoadError: HTTP.ConnectError for url = `[https://geodata.ucdavis.edu/climate/worldclim/2_1/base/wc2.1_10m_bio.zip`:](https://geodata.ucdavis.edu/climate/worldclim/2_1/base/wc2.1_10m_bio.zip%60:) TimeoutException: try_with_timeout timed out after 30.0 seconds
```

We can make robots that code better than me but the worldclim servers are still down every other day

PR #1018, opened 2025-09-15
[WIP] Non breaking geometry lookup

asinghvi17:
Add the geometry lookup things, and predicate zonal on all of this


asinghvi17:
@felixcremer this is the successor

PR #1023, opened 2025-10-05
add `emptyval` keyword to `zonal`

rafaqz:
This PR adds an `emptyval` keyword which specifies what to return if the reduced area is empty.

@kadyb this should mostly solve your problem, although not sure this is the best way to do it.

TODO: allow a `NamedTuple` of `emptyval` for `RasterStack`


asinghvi17 (approved):
LGTM assuming isempty is fast on a SkipMissingVal

rafaqz:
No, `missing` or `NaN` init will always return missing or NaN.

@asinghvi17 `isempty isn't fast if the area is mostly empty. Otherwise we could check and throw a nicer error we no `emptyval` is given.

PR #1027, opened 2025-10-14
Do not use `Base.get_extension`

### LINKED ISSUE #1025: Use of `Base.get_extension`
Was just reading the community newsletter and came across https://github.com/JuliaLang/julia/pull/59593 - which says 

> It is not allowed for packages to call Base.get_extension

`Base.get_extension` is exactly how we check this - should we replace it with something dispatch-based?

tiemvanderdeure:
Drop the use of `Base.get_extension` in favour of an is_loaded function that is overloaded inside of each extension

Closes #1025

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -15,8 +15,10 @@ permissions:
   statuses: write
 
 concurrency:
-  group: pages
-  cancel-in-progress: true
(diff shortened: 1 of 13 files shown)


asinghvi17:
https://github.com/tiemvanderdeure/Rasters.jl/pull/4

PR #1030, opened 2025-10-16
fix threadid for 1.12

### LINKED ISSUE #1026: Threaded rasterize is broken in CI
See https://github.com/rafaqz/Rasters.jl/actions/runs/18469560661/job/52619784381?pr=1024 and the discussion in https://github.com/rafaqz/Rasters.jl/pull/1024. 

We use `threadid()` in a way we probably should not inside of `_run`, but I  haven't been able to reproduce this locally yet. I'm also not entirely sure how many threads CI runs on (and if it has an interactive thread, like was introduced in 1.12?)

We probably need to rewrite `_run`

tiemvanderdeure:
Closes #1026

This does all feel like a silly breakage - even though I know we're not supposed to use threadid like this

Let's see if tests pass now

### DIFF
--- a/src/methods/coverage.jl
+++ b/src/methods/coverage.jl
@@ -300,7 +300,7 @@ function _sum_coverage!(A::AbstractRaster, geoms, buffers;
         close(allbuffers)
         _buffers = collect(allbuffers)
         _do_broadcast!(+, A, getindex.(_buffers, :coveragebuffer)...)
-        return sum(first.(getindex.(_buffers, :missed_pixels)))
+        return sum(x -> first(x.missed_pixels), _buffers)
 
     else
         _do_broadcast!(+, A, allbuffers.coveragebuffer)


tiemvanderdeure:
I do kind of want to make sure I didn't break anything for 1.10. Is there any reason why we're only running CI on Julia 1?

test/sources/gdal.jl
```diff
@@ -43,7 +43,7 @@ gdalpath = maybedownload(url)
 
         @test parent(gdalarray) isa Matrix{UInt8}
         @test parent(nomissing_array) isa Matrix{UInt8}
-        @test parent(missing_array) isa Base.ReshapedArray{Union{Missing,UInt8}}
+        @test parent(missing_array) isa Matrix{Union{Missing,UInt8}}
```
tiemvanderdeure: I'm don't really understand why this changed - but it seems fine that the parent array of an eagerly loaded raster is a Matrix?

rafaqz:
> think that with these changes we can get rid of the :static flag which should speed things up a lot

Static is usually the fastest. This will almost certainly be slower, channel is pretty heavy unless there is a lot of work fir each thread.

The reason for this switch in base is that it's composable/nestable, not faster. :static isn't nestable. But thats almost never a problem for us except after Julia bug after Ctl-C triggering nested static warnings.

src/methods/burning/geometry.jl
```diff
@@ -82,7 +83,7 @@ function _burn_geometry!(B::AbstractRaster, ::GI.AbstractGeometryTrait, geom;
     kw...
 )::Bool
     hasburned = false
-    GI.npoint(geom) > 0 || return hasburned
+    GI.isempty(geom) && return hasburned
```
tiemvanderdeure: @asinghvi17 can you confirm this is correct? I profiled and found npoint allocated a lot (at least for gdal geometries). We don't need to count the actual number of points here :)

tiemvanderdeure:
Okay this was expected but whole `_maybe_channel` business of course is giving problematic type inference issues... I was hoping that the union would resolve itself because inside of run there shouldn't be any type instability anymore, but it's a bit more tricky than that.

src/methods/coverage.jl
```diff
@@ -95,16 +95,14 @@ function _coverage!(A::AbstractRaster, ::GI.AbstractGeometryTrait, geom, r; scal
...
     buffers = (
-        # We always need a vector of threads, but just
-        allocs = _burning_allocs(A; nthreads=n, threaded=true),
-        linebuffer = [_init_bools(A, BitArray; missingval=false) for _ in 1:n],
-        centerbuffer = [_init_bools(A, BitArray; missingval=false) for _ in 1:n],
-        block_crossings = [[Vector{Float64}(undef, 0) for _ in 1:scale] for _ in 1:n],
-        burnstatus=[fill(BurnStatus(), scale) for _ in 1:n],
-        subbuffer = [fill!(Array{Bool}(undef, scale, scale), false) for _ in 1:n],
-        ncrossings = [fill(0, scale) for _ in 1:n],
+        allocs = _burning_allocs(A; nthreads=1, threaded=false), # allocate just one for now
```
tiemvanderdeure: This is giving problems later on... We could just remove the _maybe_channel part of _burning_allocs

docs/src/tutorials/plotting.md
asinghvi17: I fixed this in my PR so you can discard these changes, if you want

src/utils.jl
```diff
@@ -476,6 +476,29 @@ function _run(f, range::OrdinalRange, threaded::Bool, progress::Bool, desc::Stri
...
+        f(x) # do your work
+    finally
+        put!(resource, x) # put resource back
+    end
+end
+with_resource(f, a) = f(a)
+
+function _maybe_channel(x, threaded, n)
+    if threaded
+        ch = Channel{eltype(x)}(n)
```
asinghvi17: Shouldn't this be `Channel{typeof(x)}` instead?  Since we are deepcopying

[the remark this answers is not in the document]
asinghvi17: This can sometimes help the compiler out and make it specialize

Also might be worth making sure that threaded is converted to BoolsAsTypes

tiemvanderdeure: yes you're right this is my mistake!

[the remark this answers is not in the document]
tiemvanderdeure: Right now we're just passing threaded around as a bool. But you might be right that we might need to use `_True` and `_False`

src/methods/coverage.jl
tiemvanderdeure: this wasn't actually the problem

tiemvanderdeure:
I've added a function barrier in union_coverage which fixed the type instability, though I still don't 100% understand where it came from. I think long term we should still consider to use `_True` and `_False` for threaded instead of just bools - it will probably make all of our code much more robust to type instability popping up every time we change something. We can dispatch on both `Bool`, `_True`, and `_False` in functions like `_maybe_channel`, so that we don't have to switch over all methods in 1 go.

The new benchmarks are
```
Benchmark results (threaded = false):
--------------------------------------------------
Operation               │ Time (s) │ Memory (GB)
--------------------------------------------------
rasterize (sum)        │    0.201 │       0.03
rasterize (first)      │    2.640 │       0.25
coverage (union)       │    7.833 │       0.08
coverage (sum)         │    8.318 │       0.07
extract                │    0.241 │       0.07
extract (skipmissing)  │    0.311 │       0.07
--------------------------------------------------

Benchmark results (threaded = true):
--------------------------------------------------
Operation               │ Time (s) │ Memory (GB)
--------------------------------------------------
rasterize (sum)        │    1.026 │       0.25
rasterize (first)      │    0.648 │       0.25
coverage (union)       │    1.874 │       0.28
coverage (sum)         │    2.360 │       0.20
extract                │    0.104 │       0.07
extract (skipmissing)  │    0.065 │       0.07
--------------------------------------------------
```

tiemvanderdeure:
I don't know how reliable these benchmarks are but I'm quite sure that this is not a major performance regression, and in some cases it probably gives better performance. If tests pass I think we should just merge this now so we can merge all the other PRs. We can always come back to it later.

EDIT: We should probably also make a release because threaded operations are just broken right now for anyone on julia 1.12

src/methods/coverage.jl
```diff
@@ -259,31 +239,75 @@ function _union_coverage!(A::AbstractRaster, geom;
...
         return nothing
     end
-    _do_broadcast!(+, A, coveragebuffers...)
     _set_burnchecks(burnchecks, metadata(A), verbose)
-    return sum(missed_pixels)
+    if threaded
+        close(allbuffers)
+        _buffers = collect(allbuffers)
+        _do_broadcast!(+, A, getindex.(_buffers, :coveragebuffer)...)
+        return sum(first.(getindex.(_buffers, :missed_pixels)))
```
asinghvi17:
```suggestion
        return sum(x -> first(x.missed_pixels), _buffers)
```

To cut down on those allocations

asinghvi17 (approved):

PR #1033, opened 2025-10-24
Actually support user-defined `op` in `rasterize`

asinghvi17:
Even with providing `eltype` and `init` there were certain things trying to be computed that shouldn't have.  This fixes that. 

Also adds a test about this and streamlines CI + adds a `lts` job, as well as a doc example

Here's the MWE from slack:
```julia
using Rasters, ArchGDAL
import GeoInterface as GI
using GADM
countries_fc = GADM.get("USA"; depth = 1)

ext = Rasters.Extent(X = (-180, 180), Y = (-90, 90))
width, height = 720, 360

target = Rasters.create(
    Vector{Int},
    ext;
    size = (width, height),
    fill = Int[],
    missingval = Int[],
)

rasterize!(
    target,
    countries_fc;
    fill = [[i] for i in 1:GI.nfeature(countries_fc)],
    missingval = Int[],
    init = Int[],
    op = vcat,
    boundary = :touches,
    eltype = Vector{Int},
    progress = true,
)
```

There seems to be another issue here though, that apparently no two polygons in the dataset share the same cell?!  Which seems...incorrect

### DIFF
--- a/docs/src/tutorials/crazy_rasterization.md
+++ b/docs/src/tutorials/crazy_rasterization.md
@@ -6,7 +6,7 @@ A benefit of being written in pure Julia is that you can do anything you want wi
 Let's take an example.  Say you want to rasterize all the countries in the world, but 
 where two countries overlap, you want to save the indices of both countries.
 
-A simple but inefficient way to do this is to have each pixel of the raster actually be an
+A simple way to do this is to have each pixel of the raster actually be an
 array of integers, which stores the index of each geometry that touches the pixel.
 
 First, let's get some data.  This is a feature collection of all countries in the world.
--- a/docs/src/tutorials/crazy_rasterization.md
+++ b/docs/src/tutorials/crazy_rasterization.md
@@ -8,7 +8,6 @@ where two countries overlap, you want to save the indices of both countries.
 
 A simple but inefficient way to do this is to have each pixel of the raster actually be an
 array of integers, which stores the index of each geometry that touches the pixel.
-can control exactly what you want to do.
 
 First, let's get some data.  This is a feature collection of all countries in the world.
 
--- a/docs/src/tutorials/crazy_rasterization.md
+++ b/docs/src/tutorials/crazy_rasterization.md
@@ -30,7 +30,7 @@ using Rasters
 ras = rasterize(
     countries;
     op = vcat, 
-    fill = [[i] for i in 1:length(countries.geometry)],
+    fill = [[i] for i in 1:length(countries.geometry)], # By default a Vector with just the current country id
     boundary = :touches, # this one is just for the plot to look good
     res = 0.5, # half-degree resolutoin
     # Below are the kwargs you have to provide if using a
--- a/docs/src/tutorials/crazy_rasterization.md
+++ b/docs/src/tutorials/crazy_rasterization.md
@@ -1,0 +1,54 @@
+# Rasterize with crazy things
+
+A benefit of being written in pure Julia is that you can do anything you want with 
+`rasterize` - you can rasterize to any type or combination of types that you can get to work.
+
+Let's take an example.  Say you want to rasterize all the countries in the world, but 
+where two countries overlap, you want to save the indices of both countries.
+
+A simple but inefficient way to do this is to have each pixel of the raster actually be an
+array of integers, which stores the index of each geometry that touches the pixel.  Then you
+can control exactly what you want to do.
+
+First, let's get some data.  This is a feature collection of all countries in the world.
+
+```@example crazy
+using NaturalEarth: naturalearth
+using CairoMakie # plotting
+
+countries = naturalearth("admin_0_countries", 10)
+poly(countries.geometry; strokewidth = 1)
+```
+
+Next, we can rasterize it.  Note that if you're going beyond the standard datatypes like `Int`,
+`Float64`, etc., you will need to specify some extra kwargs to `rasterize`, so Rasters doesn't
+have to guess what you want. 
+
+```@example crazy
+using Rasters
+
+ras = rasterize(
+    countries;
+    op = vcat, 
+    fill = [[i] for i in 1:length(countries.geometry)],
+    boundary = :touches, # this one is just for the plot to look good
+    res = 0.5, # half-degree resolutoin
+    # Below are the kwargs you have to provide if using a
+    # custom type that might not have e.g. `zero` defined
+    eltype = Vector{Int},
+    missingval = Int[], # empty vector of Int
+    init = Int[],       # empty vector of Int
+    progress = false, # hide
+)
+```
+
+You can also see that this has a _lot_ of unique values:
+```@example crazy
+length(unique(ras))
+```
+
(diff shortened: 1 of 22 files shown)


src/methods/rasterize.jl
```diff
@@ -120,8 +120,12 @@ function Rasterizer(geom, fill, fillitr;
...
         isnothing(reducer) && isnothing(op) && !(fill isa Function) && throw(ArgumentError("either reducer, op or fill must be a function"))
     end
- 
-    op = _reduce_op(reducer)
+
+    op = if isnothing(op)
+        _reduce_op(reducer)
+    else
+        op
+    end
```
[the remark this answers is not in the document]
tiemvanderdeure: This actually fixes the problem with op becoming `nothing`. Will need a test!

asinghvi17:
Ok so apparently `rasterize!` gives me a different result than `rasterize`.  That's also not good

asinghvi17:
Here's the usecase MWE:

```
using Rasters, ArchGDAL
import GeoInterface as GI
using GADM
countries_fc = GADM.get("USA"; depth = 1)

ext = Rasters.Extent(X = (-180, 180), Y = (-90, 90))
width, height = 720, 360

target = Rasters.create(
    Vector{Int},
    ext;
    size = (width, height),
    fill = Int[],
    missingval = Int[],
)

bang = rasterize!(
    target,
    countries_fc;
    fill = [[i] for i in 1:GI.nfeature(countries_fc)],
    missingval = Int[],
    init = Int[],
    op = vcat,
    boundary = :touches,
    eltype = Vector{Int},
    progress = true,
)

nobang = rasterize(
    countries_fc;
    fill = [[i] for i in 1:GI.nfeature(countries_fc)],
    missingval = Int[],
    init = Int[],
    op = vcat,
    boundary = :touches,
    eltype = Vector{Int},
    size = (width, height),
    to = ext,
    progress = true,
)

unique(bang)
unique(nobang)

```

rafaqz:
Lets put these point/interval discussion somewhere else it will get lost when I merge this

docs/src/tutorials/crazy_rasterization.md
```diff
@@ -0,0 +1,54 @@
...
+`Float64`, etc., you will need to specify some extra kwargs to `rasterize`, so Rasters doesn't
+have to guess what you want. 
+
+```@example crazy
+using Rasters
+
+ras = rasterize(
+    countries;
+    op = vcat, 
+    fill = [[i] for i in 1:length(countries.geometry)],
```
rafaqz:
```suggestion
    fill = [[i] for i in 1:length(countries.geometry)], # By default a Vector with just the current country id
```

docs/src/tutorials/crazy_rasterization.md
```diff
@@ -0,0 +1,54 @@
+# Rasterize with crazy things
+
+A benefit of being written in pure Julia is that you can do anything you want with 
+`rasterize` - you can rasterize to any type or combination of types that you can get to work.
+
+Let's take an example.  Say you want to rasterize all the countries in the world, but 
+where two countries overlap, you want to save the indices of both countries.
+
+A simple but inefficient way to do this is to have each pixel of the raster actually be an
+array of integers, which stores the index of each geometry that touches the pixel.  Then you
```
rafaqz:
```suggestion
array of integers, which stores the index of each geometry that touches the pixel.
```

docs/src/tutorials/crazy_rasterization.md
```diff
@@ -0,0 +1,54 @@
...
+can control exactly what you want to do.
```
rafaqz:
```suggestion
Then we can use `rasterize` to add or remove countries, here using `vcat` to simply join merge vectors of country numbers.
```

docs/src/tutorials/crazy_rasterization.md
```diff
@@ -0,0 +1,54 @@
+# Rasterize with crazy things
+
+A benefit of being written in pure Julia is that you can do anything you want with 
+`rasterize` - you can rasterize to any type or combination of types that you can get to work.
+
+Let's take an example.  Say you want to rasterize all the countries in the world, but 
+where two countries overlap, you want to save the indices of both countries.
+
+A simple but inefficient way to do this is to have each pixel of the raster actually be an
```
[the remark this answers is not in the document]
rafaqz: We are just showing that you can hack, not the perfect way to hack

asinghvi17:
Hmm `sample` tests are failing on `convert`.   Seems unrelated to this PR but maybe that should be `GI.convert`?

tiemvanderdeure:
I think it's a great idea to add a rasterization tutorial - but I would prefer if it started with some more normal example and then we can have a crazy rasterization thing in the end to really show off what is possible. But since we don't have the common use cases documented as a tutorial it seems weird to add a tutorial with something so niche alone

rafaqz:
Do I need to fix this bug?

tiemvanderdeure:
> I have no clue why `sample` might be failing - didn't touch that code! And a brief look shows it doesn't seem to rely on `rasterize` at all.

It must have something to do with the changes to CI you made. I've just run Rasters test locally and no issues. My guess is we hvae  a bug specific to the lts

asinghvi17:
That seems a bit odd.  Let's see what tests on v1 give us, though they'll probably fail because of the 1.12 issue with threadid...

But I guess we just haven't been tracking lts, the CI setup here hadn't been updated in a couple years.  I also stopped the fail-fast thing, it doesn't really help when tracking down julia version issues

tiemvanderdeure:
So looking at the tests, 1.12 fails on threadid as expected, 1.11 fails on a floating point issue in the reproject tests that I already fixed in #1030, and lts fails on sample - which we should investigate.

asinghvi17:
If #1030 is working can we get it in now and then use that as a basis for the rest of the PRs?  That way, at least CI works.  cc @rafaqz 

We can also move the CI changes from here to there

tiemvanderdeure:
> Ok this is very weird...1.12 ci is passing but 1.11 and lts are not?!

Not all that weird - since previously we only tested on 1.12 and made tests pass their.

The 1.11 issue is this test https://github.com/rafaqz/Rasters.jl/blob/bf04129a9c32eb9b088fbc3bcf6d526cffd00969/test/sources/gdal.jl#L46

Reproduced it like this:
```julia
include(joinpath(dirname(pathof(Rasters)), "../test/test_utils.jl"))
url = "https://download.osgeo.org/geotiff/samples/gdal_eg/cea.tif"
gdalpath = maybedownload(url)
Raster(gdalpath; missingval=missing) |> parent |> typeof
```

On 1.11 this is `Base.ReshapedArray{Union{Missing, UInt8}, 2, Array{Union{Missing, UInt8}, 3}, Tuple{}}` and on 1.12 it is `Matrix{Union{Missing, UInt8}}`. I have no clue why though. I "fixed" this test to make it pass - but maybe we should just test that the parent is an `AbstractArray{Union{Missing, UInt8}}`

On 1.10 it fails because of this sample issue which we'll have to look into.

PR #1036, opened 2025-10-30
Update CommonDataModel compat version to 0.4

alex-s-gardner:


rafaqz:
~~Do we need to update any of the XDatasets packages at the same time?~~

Seems like we dont

rafaqz:
Right I missed that, it doesnt make sense

alex-s-gardner:
I got a conflict if I loaded NCDatatsets before Rasters.. hence the bump

Issue #1032, opened 2025-10-24
threadsafe operations are handled as if they were not threadsafe


tiemvanderdeure:
As I mentioned in https://github.com/rafaqz/Rasters.jl/pull/1030, `rasterize` is handling some threadsafe functions as if they were not. Functions like `sum` and `minimum` are converted to `Base.add_sum` and `min` somewhere - but these functions have no dispatch on `_is_op_threadsafe`, resulting in way worse performance for threaded rasterize!

For example:
```julia
using Rasters, NaturalEarth, Rasters.Lookups

# Load world countries and rivers
countries = naturalearth("10m_admin_0_countries")
# Create a global grid
ras = Rasters.create(
    Float64, Extent(X=(-180, 180), Y=(-90, 90)); 
    res = 1/12, sampling=(X=Intervals(Start()), Y=Intervals(Start()))
)

rasterize(sum, countries; to = ras, fill = 1.0, threaded = true, verbose = false, progress = false); #precompile
@time rasterize(sum, countries; to = ras, fill = 1.0, threaded = true);
Rasters._is_op_threadsafe(::(typeof(Base.add_sum))) = true # base add is threadsafe!!
rasterize(sum, countries; to = ras, fill = 1.0, threaded = true, verbose = false, progress = false);
@time rasterize(sum, countries; to = ras, fill = 1.0, threaded = true);
```

Returns
```
┌ Warning: if `op` is not threadsafe, `threaded=true` may be slower than `threaded=false`
└ @ Rasters C:\Users\tsh371\.julia\dev\Rasters\src\methods\rasterize.jl:621
Reducing... 100%|██████████████████████████████████████████████████| Time: 0:00:01
  1.258277 seconds (12.51 M allocations: 1017.519 MiB, 12.71% gc time, 20 lock conflicts)
[ Info: 20 geometries did not affect any pixels. See `metadata(raster)["missed_geometries"]` for a vector of misses
  0.085330 seconds (143.39 k allocations: 108.449 MiB, 8.73% gc time, 5 lock conflicts)
```

So this is just nuking the performance.

I guess the easiest fix is just to add these methods to `_is_op_threadsafe` - but maybe we want to take a closer look at the code to avoid this in the future?

asinghvi17:
> CommutativeOps

is that a proposed package name I hear?  :D

rafaqz:
No `Base.PermutedDimsArrays.CommutativeOps` ;)

[...]

PR #1039, opened 2025-11-10
make sure threadsafe rasterization ops are efficient

tiemvanderdeure:
closes https://github.com/rafaqz/Rasters.jl/issues/1032

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,7 +20,7 @@ end
 flush_info_and_warnings()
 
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
But that can just get deleted probably and maybe if someone can file an issue about that, I can get around to looking at it soonish

tiemvanderdeure:
I think we can merge this? The documentation issue will be fixed whenever the Makie folks bump ComputePipeline

Issue #1042, opened 2025-11-14
rasterize into NetCDF file gives cryptic error if NCDataset is not loaded


felixcremer:
When I am trying to rasterize into a netcdf file I get the following error and this is not really clear why this is not working. For a .tif file I get a better error message and also for loading a NetCDF. 
The same happens for Zarr files.

```julia
using Rasters
julia> pointvec = [
           (-20.0, 30.0),
           (-20.0, 10.0),
           (0.0, 10.0),
           (0.0, 30.0),
           (-20.0, 30.0),
       ]
julia> rasterize(count, pointvec, filename="/tmp/rasterizetest.nc", size=10)
ERROR: ArgumentError: `sourceconstructor` is not defined for Rasters.NCDsource
Stacktrace:
  [1] sourceconstructor(s::Type{Rasters.NCDsource})
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/sources/sources.jl:77
  [2] sourceconstructor(source::Rasters.NCDsource)
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/sources/sources.jl:76
  [3] write(filename::String, source::Rasters.NCDsource, A::Raster{…}; append::Bool, force::Bool, kw::@Kwargs{…})
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/sources/commondatamodel.jl:475
  [4] write(::Function, ::String, ::Vararg{…}; kw::@Kwargs{…})
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/write.jl:207
  [5] write(::Function, ::Function, ::Vararg{…}; kw::@Kwargs{…})
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/write.jl:207
  [6] create(filename::String, source::Rasters.NCDsource, ::Type{…}, dims::Tuple{…}; name::Symbol, missingval::Int64, fill::Int64, metadata::Metadata{…}, chunks::Rasters.NoKW, scale::Rasters.NoKW, offset::Rasters.NoKW, dropband::Bool, lazy::Bool, verbose::Bool, force::Bool, coerce::Rasters.NoKW, f::Rasters.var"#625#626"{…}, kw::@Kwargs{})
    @ Rasters ~/.julia/packages/Rasters/G6rYV/src/create.jl:301
[...]

PR #1043, opened 2025-11-14
more BackendException less error()

tiemvanderdeure:
closes https://github.com/rafaqz/Rasters.jl/issues/1042

### DIFF
--- a/src/sources/sources.jl
+++ b/src/sources/sources.jl
@@ -74,8 +74,7 @@ end
 
 # Dataset constructor from `Source`
 sourceconstructor(source::Source) = sourceconstructor(typeof(source))
-sourceconstructor(s::Type{<:Source}) = 
-    throw(ArgumentError("`sourceconstructor` is not defined for $s"))
+sourceconstructor(s::Type{<:Source}) = throw(BackendException(s()))
 # Function to check filename
 checkfilename(s::Source, filename) = throw(BackendException(s))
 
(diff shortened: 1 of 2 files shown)


src/sources/sources.jl
```diff
@@ -74,8 +74,7 @@ end
 
 # Dataset constructor from `Source`
 sourceconstructor(source::Source) = sourceconstructor(typeof(source))
```
tiemvanderdeure: Actually not sure why `sourceconstructor` dispatches on the types?

src/sources/sources.jl
```diff
@@ -74,8 +74,7 @@ end
...
-sourceconstructor(s::Type{<:Source}) = 
-    throw(ArgumentError("`sourceconstructor` is not defined for $s"))
+sourceconstructor(s::Type{<:Source}) = throw(BackendException(s()))
```
felixcremer: Currently writing is not implemented yet for Zarr so it should be special cased to throw an informative error that it is not implemented yet.

tiemvanderdeure: What error do you currently get - before and after loading ZarrDatasets?

tiemvanderdeure: I've never really used zarr but it looks like a zarr file? It's a folder called .zarr and I can also read it back in. But you can try to run that code with `using ZarrDatasets`

tiemvanderdeure:
Good to go? @rafaqz

src/sources/sources.jl
rafaqz: Me either 😆

rafaqz (approved):

Issue #1045, opened 2025-11-20
resample flips the data upside down


tiemvanderdeure:
Seems bad?? Or am I missing something obvious here?

```julia
using Rasters, Rasters.Lookups, GLMakie
import GADM, ArchGDAL

ras = rand(X(8:0.05:15), Y(58:-0.05:54))

# denmark boundaries
dk = GADM.get("DNK")

# rasterize
dkras = rasterize(first, dk; res = 0.1, fill = 1, missingval = 0)

# resample
dkras_res = resample(dkras; to = dims(ras, (X,Y)))

# plot
fig, ax, pl = plot(dkras)
ax2 = Axis(fig[1, 2]; title = "Resampled")
plot!(ax2, dkras_res)
fig

```

<img width="628" height="516" alt="Image" src="https://github.com/user-attachments/assets/433230c4-bba2-4c76-8fd8-65df7c44d256" />

rafaqz:
Ugh yeah gdal does that. We just have to fix the lookup, maybe that isnt tested

PR #1046, opened 2025-11-22
Fix flipped outputs from resample

tiemvanderdeure:
closes https://github.com/rafaqz/Rasters.jl/issues/1045

I figured out this occured when resampling to dimensions that have different order than the dimensions of the raster that was being resampled. A call to `reorder` should fix that.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,7 +20,7 @@ end
 flush_info_and_warnings()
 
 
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
Not sure what is going on with all those docstrings... It reports there are a bunch of downloading messages in them that aren't supposed to be there

PR #1067, opened 2026-03-09
[Breaking] Set missingval of count rasterize to non-zero after rasterisation

felixcremer:
Before this change the missingval of a count rasterize result was set to zero.
I find this confusing, because I would have expected that zero is a valid result of a count rasterisation. This change therefore sets the missingval metadata after the rasterisation to the default missingval for the eltype and filename or to the provided keyword value. 

I tried setting it in the rasterize call itself, but then it changed the valid zero values to missing. 
This adds a few tests for this behaviour.
This is a breaking change because the default missingval of a rasterize result with count changes.


rafaqz:
This seems fine but tests need to pass

felixcremer:
I had to change the tests slightly, because the mask operation doesn't work with the rasterized count anymore because zero is not the missingval anymore and I hope I translated the idea of the test correctly.

I also allowed scalar indexing for the rasterization to disk of a point vector because this is inherently scalar.

This should be good to go now.

rafaqz:
@felixcremer it would be nice to remove that comment about scalar rasterization! we need a chunk table to put points in then we can do it chunk-by-chunk on disk. Polgyons etc too if we store the scan state at the boundary.

PR #1070, opened 2026-04-20
Range extremes computation leads to numerical precision bugs

dnstudent:
While working with a high-resolution GeoTIFF DEM (GMTED 7.5"), I noticed that the coordinate steps retrieved via RasterStack("dem.tif", lazy=true) exhibit slight floating-point imprecision in the final decimal places. Interestingly, both gdalinfo and ArchGDAL report these values correctly.

Upon investigation, I identified a snippet in the RastersArchGDALExt extension that likely introduces these numerical deltas. Currently, the dimension range is recomputed using start, end, and count. Because the end value is derived from a multiplication step, it introduces small errors compared to using start, step, and count—which are the raw input factors.

https://github.com/rafaqz/Rasters.jl/blob/3ba72770c1d8aab8edb1b124e151baeb1d660ef2/ext/RastersArchGDALExt/gdal_source.jl#L161-L179

I have prepared a pull request to address this, along with two other instances of the same pattern found within the codebase.

Note on Coordinate Offsets:
I have proceeded under the assumption that ArchGDAL consistently provides the Upper Left corner coordinates (as per the GDAL geotransform standard).

### DIFF
--- a/ext/RastersArchGDALExt/gdal_source.jl
+++ b/ext/RastersArchGDALExt/gdal_source.jl
@@ -158,32 +158,20 @@ function RA._dims(raster::AG.RasterDataset, crs=nokw, mappedcrs=nokw)
         xstep = gt[GDAL_WE_RES]
         ystep = gt[GDAL_NS_RES] # Usually a negative number
         # Get min, max and sampling depending on AREA_OR_POINT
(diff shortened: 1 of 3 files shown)


rafaqz:
I think this is actually correct - you just need to fix the test failures! 

Just swap the range type.

PR #1075, opened 2026-05-25
Stable ranges

rafaqz:
@dnstudent I fixed the test problem here, thanks for the PR

@tiemvanderdeure can I get a sanity check on this, but it seems like the right move to me. Will go in the breaking branch to handle the slight change.

Edit:  the solution ended up being more complicated than #1070 - we need anchored ranges that use the `offset` field of `StepRangeLen`. Should be more accurate


test/aggregate.jl
```diff
@@ -56,8 +56,8 @@ series = RasterSeries([stack1, stack2], (Ti(dates),))
     @test lookup(aglat) == LinRange(-10.0, 5.0, 2)
     disaglat = disaggregate(Start(), aglat, 3)
     # The last item is lost due to rounding in `aggregate`
-    @test lookup(disaglat) != lookup(dimz[2])
-    @test parent(lookup(disaglat)) === LinRange(-10.0, 15.0, 6)
+    @test index(disaglat) != index(dimz[2])
+    @test index(disaglat) === range(; start=-10.0, step=5.0, length=6)
```
rafaqz:
```suggestion
    @test lookup(disaglat) != lookup(dimz[2])
    @test parent(lookup((disaglat)) === range(; start=-10.0, step=5.0, length=6)
```

test/aggregate.jl
```diff
@@ -57,7 +57,7 @@ series = RasterSeries([stack1, stack2], (Ti(dates),))
     disaglat = disaggregate(Start(), aglat, 3)
     # The last item is lost due to rounding in `aggregate`
     @test lookup(disaglat) != lookup(dimz[2])
-    @test parent(lookup(disaglat)) === LinRange(-10.0, 15.0, 6)
+    @test parent(lookup((disaglat)) === range(; start=-10.0, step=5.0, length=6)
```
rafaqz:
```suggestion
    @test parent(lookup(disaglat)) === range(; start=-10.0, step=5.0, length=6)
```

src/sources/grd.jl
```diff
@@ -79,8 +79,8 @@ function _dims(A::RasterDiskArray{GRDsource}, crs=nokw, mappedcrs=nokw)
     # Not fully implemented yet
     xy_metadata = _metadatadict(GRDsource())
 
-    xindex = range(; start=xbounds[1], stop=xbounds[2] - xspan, length=xsize)
-    yindex = range(; start=ybounds[2] + yspan, stop=ybounds[1], length=ysize)
+    xindex = range(; start=xbounds[1], step=xspan, length=xsize)
+    yindex = range(; start=ybounds[2] + yspan, step=yspan, length=ysize)
```
tiemvanderdeure: Can't see what this change solves? xspan and yspan are still computed from the bounds and size, so we'd run into the same floating point issues? If .grd files only report their extent (but not the step), then I don't think it'd be possible to completely avoid floating point issues

tiemvanderdeure (approved):

tiemvanderdeure:
Seems the last commit reversed most of @dnstudent 's changes! Probably not intentional?

rafaqz:
Yeah just claude "fixing" things.

Problem is the change introduces a lot of bugs. I'll start again.

src/sources/grd.jl
rafaqz: We know the step exactly from the geotransform, but the stop we calculate. The means the step can very slightly vary. Its actually bad, means `cat` can fail on non-matching step size.

tiemvanderdeure: Not for .grd I think. The step is defined just a few lines above as `xspan = (xbounds[2] - xbounds[1]) / xsize`, which surely can introduce floating point error in some cases. So seems like we know extent and size, but not stop or step.

Maybe we can avoid some of it with `xindex = range(; start = xbounds[1], stop = xbounds[2], length = xsize+1)[1:end-1]` because of TwicePrecision

rafaqz: There is a best of both worlds option using the offset in StepRangeLen centered on zero. Will push soon

rafaqz:
@dnstudent and @tiemvanderdeure I've fixed this with a more complicated approach - `anchored_range`. It uses a fixed step but also anchores the range start position at the zero crossing (most accurate place and standardised for ===). Then it uses the TwicePrecision mechanics to make sure we dont get fp errors on reindex later on. 

Note that this *doesn't* exactly match GDAL, because its more accurate than GDAL due to using `TwicePrecision` more cleanly. GDAL exactly matches the simple `range` approach - but it fails some tests we had that were more accurate with `LinRage` - as `LinRange` always hits the exact end point.

rafaqz:
Proof, to run in root dir of this branch or Rasters. `anchored_range` matches BigFloat 256 exactly, range doesn't.

```julia
using Rasters, ArchGDAL

ds = ArchGDAL.readraster("test/data/cea.tif")
gt = ArchGDAL.getgeotransform(ds)
start_x, step_x = gt[1], gt[2]
N = 514

r_plain = range(; start=start_x, step=step_x, length=N)
r_anchored = Rasters.anchored_range(start_x, step_x, N)

# Compute `start + (i-1)*step` in arbitrary-precision BigFloat and round to
# the nearest Float64. This is the "ideal" Float64 answer.
setprecision(BigFloat, 256)
true_vals = [Float64(BigFloat(start_x) + (i - 1) * BigFloat(step_x)) for i in 1:N]

plain_err = [abs(r_plain[i] - true_vals[i]) for i in 1:N]
anchored_err = [abs(r_anchored[i] - true_vals[i]) for i in 1:N]

n_plain_better = sum(plain_err .< anchored_err)
n_anchored_better = sum(anchored_err .< plain_err)
n_tied = sum(plain_err .== anchored_err)
n_plain_exact = sum(iszero, plain_err)
n_anchored_exact = sum(iszero, anchored_err)

println("Range length: $N")
println("Plain `range(start, step, length)`:")
println("  max error vs BigFloat-rounded-to-Float64: ", maximum(plain_err))
println("  indices where Float64 result == ideal:    $n_plain_exact / $N")
println()
println("`anchored_range`:")
println("  max error vs BigFloat-rounded-to-Float64: ", maximum(anchored_err))
println("  indices where Float64 result == ideal:    $n_anchored_exact / $N")
println()
println("Head-to-head:")
println("  plain closer to ideal at:    $n_plain_better indices")
println("  anchored closer to ideal at: $n_anchored_better indices")
println("  tied:                        $n_tied indices")
```

```
Range length: 514
Plain `range(start, step, length)`:
  max error vs BigFloat-rounded-to-Float64: 3.637978807091713e-12
  indices where Float64 result == ideal:    305 / 514

`anchored_range`:
  max error vs BigFloat-rounded-to-Float64: 0.0
  indices where Float64 result == ideal:    514 / 514

Head-to-head:
  plain closer to ideal at:    0 indices
  anchored closer to ideal at: 209 indices
  tied:                        305 indices
```

src/methods/aggregate.jl
```diff
@@ -276,16 +276,15 @@ function disaggregate(l::Lookup, scale)
...
     intscale == 1 && return l
 
-    len = length(l) * intscale
-    step_ = step(l) / intscale
-    start = first(l) - _agoffset(l, intscale) * step_
-    stop = start + (len - 1)  * step_
-    index = LinRange(start, stop, len)
+    len = length(lookup) * intscale
+    step_ = step(lookup) / intscale
+    start = lookup[1] - _agoffset(Start(), intscale) * step_
```
rafaqz:
```suggestion
    len = length(l) * intscale
    step_ = step(l) / intscale
    start = l[1] - _agoffset(Start(), intscale) * step_
```

tiemvanderdeure:
A bit surprised that this works! When I tried making some ranges it seems they were already anchored somewhere around 0. 

Have you tried this for other files that cea.tif?

src/methods/aggregate.jl
```diff
@@ -278,14 +278,13 @@ function disaggregate(l::Lookup, scale)
 
     len = length(l) * intscale
     step_ = step(l) / intscale
-    start = first(l) - _agoffset(l, intscale) * step_
-    stop = start + (len - 1)  * step_
-    index = LinRange(start, stop, len)
+    start = l[1] - _agoffset(Start(), intscale) * step_
```
rafaqz:
```suggestion
    start = l[1] - _agoffset(l, intscale) * step_
```

rafaqz:
Really? ranges normally have `offset=1` if you use `start`, `step` and `length`?

Like they dont tweak the offset at all?

tiemvanderdeure:
> Really? ranges normally have `offset=1` if you use `start`, `step` and `length`?
> 
> Like they dont tweak the offset at all?

They seem to? Or am I missing something?

```
julia> r = range(; start = -10.5, stop = 1.5, length = 10)
-10.5:1.3333333333333333:1.5

julia> r.offset
9

julia> r = range(; start = -10.5, step = 4/3, length = 10)
-10.5:1.3333333333333333:1.5

julia> r.offset
9
```

rafaqz:
With positive starting points?

rafaqz:
```julia
julia> r = range(; start = 100000.5, step = 4/3, length = 10)
100000.5:1.3333333333333333:100012.5

julia> r.offset
1
```

tiemvanderdeure:
Has something to do with the number of digits it seems

```
julia> range(start = -142.89421391, step = 5.0, length = 614).offset
1

julia> range(start = -142.894, step = 5.0, length = 614).offset
30
```

tiemvanderdeure:
> With positive starting points?

With positive starting points anchored_range also will have an offset of 1

```
julia> Rasters.anchored_range(10000.5, 4/3, 10).offset
1
```

tiemvanderdeure:
Anyway, if it works it works, just surprising that it is possible to beat Base in constructing a range. But I don't really understand the base internals here and it wouldn't be a great idea to use any of those internals even if we could

rafaqz:
Yeah, youre right about the anchor offset. Thats actually a limitation - the space it improves isn't as big as I though.

But the differnce is the offset will more often be inside the range (specifically where it crosses zero), and doesn't force a recalculation of `ref` after slicing - thats where the tiny fp innacuracies come from. And we do slicing everywhere so it matters.

But I think there might be another fix in DD that uses TwicePrecision tricks in lookups instead then we can just use `range` here and avoid the extra fp error elsewhere.

(But this means DD PR comes first)

rafaqz:
I'm actually getting some similar errors in `cat` and `mosaic` from fp error so trying to fix it too.

rafaqz:
Maybe its worth merging this too actually it is slightly better for geospatial ranges, slightly less frequently loses.

We may need our own range type in DD to do any better than this - one that is designed for repeated slicing not introducing fp error.

```julia
 struct StableRange{T<:AbstractFloat} <: AbstractRange{T}                                                
      start::Base.TwicePrecision{T}                                                                       
      step::Base.TwicePrecision{T}                                                                        
      len::Int                                                                                            
      offset::Int                                                                                         
  end                        

  Base.getindex(r::StableRange{T}, i::Int) where T = T(r.start + (i - 1 + r.offset) * r.step)             
  
  function Base.getindex(r::StableRange, s::UnitRange{Int})                                               
      StableRange(r.start, r.step, length(s), r.offset + first(s) - 1)
  end   
  ```

So we trade some fp headroom for a range always being over the same exact values no matter how many times you slice it.

(we can be better than base at repeated slicing because they are aiming for maximum local accuracy not repeated slice accuracy)

**Comment by rafaqz** [comment]
Ok I implemented `StableRange` here. Should solve a lot of these problems.

File: src/utils.jl [hunk 6]
```diff
@@ -449,6 +449,24 @@ for op in (:+, :-)
...
 
+# Scalar broadcast. Without this, `sr .+ x` falls through to the generic
+# `AbstractRange` broadcast machinery and returns a `StepRangeLen` whose
+# values can drift by a ULP from the parent (DimensionalData's
+# `_shiftlocus` does `parent(l) .+ (step * offset)` even when `offset == 0`,
+# which would otherwise convert every `StableRange` lookup into a
+# `StepRangeLen` on every locus-preserving set call).
+for op in (:+, :-)
+    @eval function Base.broadcasted(::Base.Broadcast.DefaultArrayStyle{1}, ::typeof($op), r::StableRange{T}, x::Number) where T
+        StableRange{T}(($op)(r.start, Base.TwicePrecision{T}(T(x))), r.step, r.len, r.offset)
```
tiemvanderdeure:
```suggestion
        StableRange{T}(($op)(r.start, T(x)), r.step, r.len, r.offset)
```

I think this is already handled correctly? At least `p - sqrt(2f0) === p - Base.TwicePrecision{Float64}(Float64(sqrt(2f0)))` holds where p = Base.TwicePrecision(pi)

src/utils.jl
```diff
@@ -449,6 +449,24 @@ for op in (:+, :-)
...
 end
 
+# Scalar broadcast. Without this, `sr .+ x` falls through to the generic
+# `AbstractRange` broadcast machinery and returns a `StepRangeLen` whose
+# values can drift by a ULP from the parent (DimensionalData's
+# `_shiftlocus` does `parent(l) .+ (step * offset)` even when `offset == 0`,
+# which would otherwise convert every `StableRange` lookup into a
+# `StepRangeLen` on every locus-preserving set call).
+for op in (:+, :-)
+    @eval function Base.broadcasted(::Base.Broadcast.DefaultArrayStyle{1}, ::typeof($op), r::StableRange{T}, x::Number) where T
```
tiemvanderdeure, line 533: We need to add the opposite order as well so that `1 .+ stablerange` also returns a StableRange

tiemvanderdeure: okay it's just below - my bad!

src/utils.jl
rafaqz: Sorry Im not sure what change you are suggesting

tiemvanderdeure: No need to call `Base.TwicePrecision{T}` on `x`, the outcome is the exact same without. I think even converting to T is not needed

rafaqz: Ahh right. The deleted whitespace threw me off - I think you are removing all indentation

(I fixed it)

rafaqz: I think I need a little more evidence for this claim that its exactly the same - do you mean provably always or in one example? Most of these things *nearly* always give the same answer.

rafaqz: Ok so we dont need TicePrecision but we do need T

rafaqz:
Later we can make a little StableRanges.jl so this can be used elsewhere. I think its pretty fundamental for spatial work now I understand the difference. 

We dont want `ref` positions recalculated ever - we want all lookup positions to remain identical through infinite slicing/merging operations. Thats more important to our use cases than slightly more floating point headroom, we dont need to calculate the position of atoms.

We will need this separate from Rasters for `cat` to merge stable ranges back into the same offseted `StableRange` in DD.

PR #1065, opened 2026-03-03
Breaking: preserve extent when creating a Raster

tiemvanderdeure:
Closes #1031

This will probably break a bunch of tests (haven't totally figured out my AI Agent setup yet to fix all that)

Because of how `range` works, if res and extent are not compatible, the raster will be 1 cell too small rather than big. In most cases I don't think that's an issue but we might want to change it.

PR #1076, opened 2026-05-26
Breaking

rafaqz:


tiemvanderdeure:
I'm a bit sad we didn't get https://github.com/rafaqz/Rasters.jl/pull/1065 in.

I can give it another go this week so we could include it into the next release?

PR #1083, opened 2026-06-18
allow rasterizing with in-place op

tiemvanderdeure:
Using a vector/set as eltype in `rasterize` was possible since https://github.com/rafaqz/Rasters.jl/pull/1033, but for an `op` that modifies in-place this doesn't work because the `init` is never copied.

This PR adds that, and also some performance improvement by forcing `rasterize_points_op!` to specialize on `op`

My use case was finding out which species are present in each grid cell in a very large database of presence data. The order really does not matter here, so the eltype is Set and the op is push!

This now works:
```
n = 100_000
geoms = [(rand(), rand()) for _ in 1:n]
species = ["Species" * string(rand(1:3)) for _ in eachindex(geoms)]
@time ras_species = rasterize(geoms; res = 0.001, fill = species, op = push!, init = Set{String}(), missingval = Set{String}(), eltype = Set{String})
unique(ras_species)
```

And it takes under 0.1 second for 100 thousand points.

There is a bit of room for improvement here to not have to specify all of `init`, `missingval`, and `eltype`, but I couldn't figure out how to make that possible without breaking anything.


src/methods/rasterize.jl
```diff
@@ -769,7 +767,7 @@ function _fill_func!(fillfunc, A::RasterStack, I)
 end
 
 # Some reductions don't need sort, like sum
-@noinline function rasterize_points_op!(op, A, points_fill, missingval, init)::Bool
+@noinline function rasterize_points_op!(op::F, A, points_fill, missingval, init)::Bool where F<:Function
```
asinghvi17:
```suggestion
@noinline function rasterize_points_op!(op::O, A, points_fill, missingval, init)::Bool where O
```

could pass a callable struct too, which is not `<: Function`

src/methods/rasterize.jl
```diff
@@ -982,3 +980,15 @@ Base.@assume_effects :total _apply_op(op::F, a1, fill) where F<:Function = op(a1
...
+# propagate through NamedTuples so each cell gets its own instance.
+function _maybe_copy_init(x)
+    if x isa NamedTuple
+        return map(_maybe_copy_init, x)
+    elseif x isa AbstractArray || x isa AbstractSet || x isa Dict
+        return copy(x)
+    else
+        return x
+    end
+end
```
asinghvi17:
```suggestion
function _maybe_copy_init(x)
    if x isa NamedTuple
        return map(_maybe_copy_init, x)
    elseif Base.ismutable(x)
        return copy(x) # deepcopy?
    else # immutable
        return x
    end
end
```

It might be worth just `deepcopy`ing generally, since you could in theory have nested things, or a `struct A; a::Vector; end`.

That replaces this with:

```suggestion
_maybe_copy_init(x) = deepcopy(x)
```

asinghvi17: But deepcopy might also not be desirable for e.g. a CategoricalArray, so this way you at least leave it open for dispatch.

asinghvi17 (changes requested):

src/methods/rasterize.jl
tiemvanderdeure: Yeah I thought about this - then we don't need _maybe_copy_init at all, might as well just call `deepcopy` where needed. I thought there might be cases where deepcopy would be unnecessary and expensive, but I really cannot think of any. For number types it's a no-op

tiemvanderdeure: Okay actually just tried deepcopy, and even for Set it adds quite a bit of overhead, so let's just stick to `copy`.

For really complex types users can always just make the resulting Raster and use rasterize!

asinghvi17 (approved):

tiemvanderdeure:
TIL that a Set is not technically mutable! Checking if `x isa AbstractSet || Base.ismutable(x)` is maybe slightly clunky but I think still okay.

rafaqz (approved):
Thanks! Love how flexible rasterize is getting

src/methods/rasterize.jl
tiemvanderdeure: I changed this back to `x isa AbstractArray || x isa AbstractSet || x isa AbstractDict` after testing a bit more. Something like a String is also technically mutable, but `copy` is not defined, and `deepcopy` is expensive.

src/methods/rasterize.jl
```diff
@@ -222,11 +222,11 @@ function _get_eltype_missingval(eltype, missingval, filleltype, fillitr, init::N
 end
 function _get_eltype_missingval(known_eltype, missingval, filleltype, fillitr, init, filename, op, reducer)
     eltype = if isnothing(known_eltype)
-        fillzero = zero(filleltype)
+        fillzero = _zero_or_empty(filleltype)
         if fillitr isa Function
             promote_type(typeof(fillitr(init)), typeof(fillitr(fillzero)))
         elseif op isa Function
-            promote_type(typeof(op(init, init)), typeof(op(init, fillzero)), typeof(op(fillzero, fillzero)))
+            promote_type(typeof(op(_maybe_copy(init), fillzero)))
```
tiemvanderdeure, line 229: `op(init, init)` should never be called, so I think only `op(init, fillzero)` is really enough here

tiemvanderdeure:
Tweaked the eltype finding function a bit so that this now works without providing `eltype`, also added a test for a case where fill is a vector of `Char` and `op = *` and `init = ""`, which also works

Maybe we can expand https://rafaqz.github.io/Rasters.jl/dev/tutorials/crazy_rasterization#Rasterize-with-crazy-things a little bit. The rasterizing species names into sets was an actual use case ;)

asinghvi17 (approved):
LGTM, should we merge?

PR #1089, opened 2026-08-21
Fix multipoint extract return type

ThatcherC:
Claude found this bug and proposed this fix while I was working on a separate branch of Rasters.jl. This is a simple return type error that pops up when using multi-point geometries with `extract`. The PR adds a quick test to check that the function works as expected, and corrects the return type in `_extract_point!`.

### DIFF
--- a/test/extract.jl
+++ b/test/extract.jl
@@ -115,6 +115,21 @@ extr[2].index
     end
 end
 
+@testset "MultiPoint" begin
+    T = @NamedTuple{geometry::Union{Missing,Tuple{Float64,Float64}},test::Union{Missing,Int64}}
+    mp = GI.MultiPoint([(9.0, 0.1), (10.0, 0.2), (10.0, 0.3)])
+    # `skipmissing=false` is the default, and used to throw a TypeError here
+    @test all(extract(rast, mp) .=== T[
+        (geometry = (9.0, 0.1), test = 1)
+        (geometry = (10.0, 0.2), test = 4)
+        (geometry = (10.0, 0.3), test = missing)
+    ])
+    @test all(extract(rast, mp; skipmissing=true) .=== @NamedTuple{geometry::Tuple{Float64,Float64},test::Int64}[
+        (geometry = (9.0, 0.1), test = 1)
+        (geometry = (10.0, 0.2), test = 4)
+    ])
+end
+
 @testset "Polygons" begin
     # Extract a polygon
     T = @NamedTuple{geometry::Union{Missing,Tuple{Float64,Float64}},test::Union{Missing,Int64}}
--- a/test/extract.jl
+++ b/test/extract.jl
@@ -118,7 +118,6 @@ end
 @testset "MultiPoint" begin
     T = @NamedTuple{geometry::Union{Missing,Tuple{Float64,Float64}},test::Union{Missing,Int64}}
     mp = GI.MultiPoint([(9.0, 0.1), (10.0, 0.2), (10.0, 0.3)])
-    # `skipmissing=false` is the default, and used to throw a TypeError here
     @test all(extract(rast, mp) .=== T[
         (geometry = (9.0, 0.1), test = 1)
         (geometry = (10.0, 0.2), test = 4)
--- a/src/methods/extract.jl
+++ b/src/methods/extract.jl
@@ -515,7 +515,7 @@ end
             _maybe_add_fields(T, props, id, coords, I)
         end
     end
-    return 1
+    return true
 end
 
 # Maybe add optional fields


asinghvi17:
Hmm, should we instead error with multipoints in extract?  I think there's a reasonable expectation that you get output sized the same as the input, and if you wanted to use extract to get a feature collection, your output will not be the multipoint you got in.

I have no objections to the fix as it stands (it's quite minimal :D) but wondering if accepting MultiPoints here even makes sense.

ThatcherC:
Erroring with multipoints seems reasonable to to me! I usually just use `extract` with a point or list of points, and it'd be easy enough to get that from a Multipoint structure if someone wanted to do that

tiemvanderdeure:
extract works with all kinds of geometries, even polygons, so I don't see why it should error for multipoint. Current behaviour seems fine to me.

test/extract.jl
```diff
@@ -115,6 +115,21 @@ extr[2].index
     end
 end
 
+@testset "MultiPoint" begin
+    T = @NamedTuple{geometry::Union{Missing,Tuple{Float64,Float64}},test::Union{Missing,Int64}}
+    mp = GI.MultiPoint([(9.0, 0.1), (10.0, 0.2), (10.0, 0.3)])
+    # `skipmissing=false` is the default, and used to throw a TypeError here
```
tiemvanderdeure: Is this a remaining comment from your back-and-forth with the AI? Otherwise what does "used to throw an error" mean?

test/extract.jl
```diff
@@ -115,6 +115,21 @@ extr[2].index
...
+    @test all(extract(rast, mp) .=== T[
+        (geometry = (9.0, 0.1), test = 1)
+        (geometry = (10.0, 0.2), test = 4)
+        (geometry = (10.0, 0.3), test = missing)
+    ])
```
tiemvanderdeure, line 125:
```suggestion
    @test extract(rast, mp) == T[
        (geometry = (9.0, 0.1), test = 1)
        (geometry = (10.0, 0.2), test = 4)
        (geometry = (10.0, 0.3), test = missing)
    ]
```

Should also work, right?

test/extract.jl
```diff
@@ -115,6 +115,21 @@ extr[2].index
...
+    @test all(extract(rast, mp; skipmissing=true) .=== @NamedTuple{geometry::Tuple{Float64,Float64},test::Int64}[
```
tiemvanderdeure, line 126: Same here, no need for all and a broadcast, can just use `==`

test/extract.jl
ThatcherC: Yup - that's from an earlier version of the test that needed to be fixed. I'll drop that line.

test/extract.jl
ThatcherC: I can confirm that skipping the broadcast and just using `==` works here. I mostly did this because the `all(` + `.===` patterns is used elsewhere in the tests, like in the Polygons tests right below this. Should I switch it over to the `==` version or keep the style used in the Polygon tests?

test/extract.jl
ThatcherC: It doesn't actually. I just tried this out locally and proposed test failed, so I tried in the REPL:

```julia
julia> A = extract(rast, mp)
3-element Vector{@NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}}:
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((9.0, 0.1), 1))
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((10.0, 0.2), 4))
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((10.0, 0.3), missing))

julia> B = T[
               (geometry = (9.0, 0.1), test = 1)
               (geometry = (10.0, 0.2), test = 4)
               (geometry = (10.0, 0.3), test = missing)
           ]
3-element Vector{@NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}}:
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((9.0, 0.1), 1))
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((10.0, 0.2), 4))
 @NamedTuple{geometry::Union{Missing, Tuple{Float64, Float64}}, test::Union{Missing, Int64}}(((10.0, 0.3), missing))

julia> A .=== B
3-element BitVector:
 1
 1
 1

julia> A == B
missing
```

In this case, I feel like keeping the test as-is makes the most sense. What do you think?

julia> A .== B
3-element Vector{Union{Missing, Bool}}:
 1
 1
  missing
```

It looks like the issue has to do with how equality is defined for `missing`:
```julia
julia> missing == missing
missing

julia> missing === missing
true
```

tiemvanderdeure: my bad you're right!

test/extract.jl
tiemvanderdeure: fair to keep the same pattern for consistency

tiemvanderdeure (approved):
Looks good I think. @asinghvi17 if you agree keeping multipoint makes sense I will merge

asinghvi17 (approved):
Sounds reasonable if we also allowed all those other things

PR #1091, opened 2026-09-02
Persist backend-specific open kwargs across lazy reopen (open_kw)

### LINKED ISSUE #1090: consolidated-metadata key issue
ZarrDatasets.jl 0.1.6 added a consolidated::Bool keyword to ZarrDataset (JuliaGeo/ZarrDatasets.jl#18) on rafaqz's request that opens a Zarr store from its .zmetadata consolidated-metadata key instead of listing the store. Changes are needed in Rasters.jl to allow RasterStack to see those consolidated keys and not forget them on reopening.

I have a WIP branch (zarr-consolidated-repro) with a two-line fix that gets RasterStack to see the keys, plus a reproducer isolating the reopening forgetting bug — happy to push it or paste the diff if useful.

mrke:
Fixes #1090.

Adds `open_kw::NamedTuple` to `Raster`/`RasterStack`, threaded through to
`FileArray`/`FileStack` so backend-specific open options (e.g. Zarr
`consolidated`) survive a lazy reopen, not just initial discovery. Kept
separate from generic `kw...` so unrecognized keywords still error as before.

    RasterStack(url; source=Zarrsource(), open_kw=(; consolidated=true))

Includes a network-free regression test (test/sources/zarr.jl) that fails
without this fix.


tiemvanderdeure:
> `RasterStack(url; source=Zarrsource(), open_kw=(; consolidated=true))`

Not sure about this syntax... There must be other cases with source-specific keywords - how do we handle those? Maybe we can just pass any other keywords to `open`?

